PR #5, opened 2018-01-10
Add tests to four methods in table.go

aimof:
Hi.

I add tests to four methods in table.go.

Thanks.

### DIFF
--- a/table_test.go
+++ b/table_test.go
@@ -36,52 +36,52 @@ func TestNewTable(t *testing.T) {
 	//testNewTable
 	table := NewTable()
 	if !reflect.DeepEqual(table.Box, NewBox()) {
-		t.Error("table.Box is wrong")
+		t.Error("table.Box is wrong\n")
 	}
 	if table.borders {
-		t.Error("table.borders is true")
+		t.Error("table.borders is true\n")
 	}
 	if !reflect.DeepEqual(table.bordersColor, tcell.ColorWhite) {
-		t.Error("table.bordersColer is wrong")
+		t.Error("table.bordersColer is wrong\n")
 	}
 	if table.separator != rune(' ') {
-		t.Error("table.separator is not ' '")
+		t.Error("table.separator is not ' '\n")
 	}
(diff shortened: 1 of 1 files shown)


rivo:
Hi,

I appreciate your contribution. I'm planning on adding tests when we get closer to a v1.0 release because there may still be slight changes pre v1.0. Especially when testing the default values set in constructors, your tests may become obsolete quickly. I'm also not sure if we should have tests for simple setters and getters. In `TestSetBorders()` you are adding 20 lines of test code for effectively one line of actual code: `t.borders = show`. My question is, will you be willing to adapt your test cases if I make changes to my functions? (The package is only a few days old, there will likely be some changes.)

What I'm looking for are test cases where a primitive is drawn, some keyboard input is simulated, and the output is compared with the expected screen output. I'm still thinking about this.

More specific feedback to your code:

- Please don't use `tcell.Color` values in your tests. There is now a new `Style` variable with default colors. So [line 44](https://github.com/rivo/tview/pull/5/files#diff-823341bb8df52b4aef4a7844b856f6feR44) should be `Styles.BorderColor` instead.
- Could you please name your test functions `Test<type name><func name>`? For example, `TestTableSetBorders()` instead of `TestSetBorders()`. I cannot rule out that there will be functions with the same name in other components. (`TestNew<type name>` should be ok.)
- Please remove the newlines at the end of `t.Error()` calls. I'm not sure why you added them but I think they are not needed.

Thanks.

aimof:
Thanks to your kind feedbuck.

I understand my PR is not good one.
If you think the PR is not needed now, please close this PR.
Else, i rewrite this PR.

I appreciate your polite response and feedbuck.

Thanks.

rivo:
Hi,

Your code is generally ok and I appreciate your effort but I'm afraid it's a bit too early so I will close it for now. I may get back to this at a later point. But if you want to contribute, I'm happy to prepare some functions for the test cases I envision.

Let me know if you're interested and I'll have something ready for you in the next few days.

Cheers!

Issue #4, opened 2018-01-10
How can I get the value of form's element?


mostafa-asg:
Hi
First of all, thanks for your great work.
In the [data entry form](https://github.com/rivo/tview/wiki/Form) sample, How can I get the values of form?  
For example in *Save* button, how can I get the value of *first name* input field?
```Go
AddButton("Save", func() {
   // I want something like this
  firstName := form.getInputField("SOME_TEXT_BOX_ID").value()
})
```

rivo:
You're right, this was missing. I don't have form element IDs so it doesn't quite work as in your example. But you now have two options:

(1) Added a change event listener to the input field in the form:

```go
form.AddInputField("First name", "", 20, nil, func(text string) { /* handle the changed text */ })
```

(2) You get the field directly (by its index) and call one its getter (requires a cast, though):

```go
inputField := form.GetElement(0).(*tview.InputField)
text := inputField.GetText()
```

mostafa-asg:
Thank you

rivo:
Please note that I've just committed a change which renames `GetElement()` to `GetFormItem()` for consistency reasons. I hope this won't cause too much trouble.

Issue #6, opened 2018-01-10
Ability to update the view from a goroutine


dpotapov:
Hi,

I'm running some parallel blocking job (like HTTP call) in a goroutine and would like to update the view once it is completed.
Since the main loop is blocked by `screen.PollEvent()` the screen is not redrawing automatically. Some key/resize event have to trigger it. Another problem with `SetFocus()` - it blocks while polling events. 

Here is example: 
```go
app := tview.NewApplication()
a := tview.NewModal().SetText("Waiting 5 seconds...").AddButtons([]string{"Ok"})
b := tview.NewModal().SetText("Done!").AddButtons([]string{"Exit"})

go func() {
	time.Sleep(5 * time.Second) // or http.Get()
	app.SetRoot(b, false).SetFocus(b).Draw()
}()

if err := app.SetRoot(a, false).SetFocus(a).Run(); err != nil {
	panic(err)
}
```

rivo:
Thanks for the hint. I just pushed a change which will fix your example. I decided not to lock while polling for events. I don't see a reason why this couldn't be removed. I hope your issues are resolved with this now. If not, let me know. (Examples always help.)

I won't merge this pull request then.

By the way, `SetRoot()` implicitly calls `SetFocus()`.

rivo:
Sorry, just saw that your change wasn't a pull request. So all good then.

dpotapov:
Thanks!

AllanWang:
@rivo sorry to bring this back from the dead, but some of your examples call `SetRoot()` as well as `SetFocus()`. See [flexbox](https://github.com/rivo/tview/blob/master/demos/flex/main.go). Maybe you'll want to update those examples?

rivo:
Right. I missed that. Should be fixed with the latest commit (c2f07f95484c7bead2bac61de12030815e2aab58). Thanks.

PR #7, opened 2018-01-11
Fix missing call SetChangedFunc() for #4

dpotapov:


rivo:
Thanks for the changes. I would like to merge this but with a few modifications:

- Please remove the changes to `Application`. (See #6 for details.)
- I don't think we need an `accept` parameter for `AddPasswordField`. But it would be useful to include a minimum length instead. Also, for consistency, I'd like to have the `changed` parameter last. So a signature like this would be good:

```go
func AddPasswordField(label, value string, fieldLength, minLength int, maskChar rune, changed func(text string)) *Form
```
(Of course with corresponding documentation.)

If you want, I can add this function myself if you take it out of the pull request.

- The changes to `InputField` are good. Could you please add a sentence to the `SetMaskCharacter()` function documentation that a value of `0` will disable masking?

dpotapov:
Apologize, I messed up with pull requests: I had to create different branches for each change.

> If you want, I can add this function myself if you take it out of the pull request.

Please go ahead and add PasswordField!

Could you please also make a fix from this commit b536e4570ea5b4f01bc1aff4e20a661fb0f13a8b?
Thanks in advance.

rivo:
Should all be done now. Let me know if you find anything else. Cheers!

PR #8, opened 2018-01-11
Add PasswordField that hides user input

dpotapov:


rivo:
I'm confused by the different pull requests. This appears to be a duplicate. For clarity, let's keep going with #7.

PR #40, opened 2018-01-23
Flex the flex-box

verdverm:
These changes enable more fine-grained control and for dynamic changes to the flex-box system.


rivo:
See #44 for discussion

PR #41, opened 2018-01-23
All events app hook

verdverm:
These changes enable all events to be hooked from outside the application.


rivo:
See #44 for discussion

PR #42, opened 2018-01-23
 Add SSNs and put them in a public AWS bucket

verdverm:
This change set adds ids to the primitives. This will be useful in enhancements and optimizations.

Meant to be internally used mainly, but useful all over the place

(I'm using it to attach handlers to an event system and look things / render an element / route to somewhere)


rivo:
See #44 for discussion

PR #43, opened 2018-01-23
Expose list item

verdverm:
To make things more accessible


rivo:
See #44 for discussion

PR #44, opened 2018-01-23
mad-sci

verdverm:
So I realized the sequence of PRs was also a sequence of commits (should have been i.d.d.)

Anyway, if you like them all, this PR covers them all.

- Expose and make accessible
  - Flex-box items
  - List
- Generalize and capture all events
- Extend primitives
  - Add `Id()` and uuid generation
  - Add Props `map[string]interface{}`
  - Add more life-cycle hooks
    - `Mount(context map[string]interface{}) error`
    - `Unmount() error`
    - `Render() error`

### DIFF
--- a/application.go
+++ b/application.go
@@ -71,6 +71,7 @@ func (a *Application) Run() error {
 		a.Unlock()
 		return err
 	}
+	a.Unlock()
 
 	// We catch panics to clean up because they mess up the terminal.
 	defer func() {
@@ -83,7 +84,6 @@ func (a *Application) Run() error {
 	}()
 
 	// Draw the screen for the first time.
-	a.Unlock()
 	a.Draw()
 
 	// Start event loop.
--- a/application.go
+++ b/application.go
@@ -109,14 +109,14 @@ func (a *Application) Run() error {
 			}
 		}
 
-		switch event := event.(type) {
+		switch evt := event.(type) {
 		case *tcell.EventKey:
 			a.RLock()
 			p := a.focus
 			a.RUnlock()
 
 			// Ctrl-C closes the application.
-			if event.Key() == tcell.KeyCtrlC {
+			if evt.Key() == tcell.KeyCtrlC {
 				a.Stop()
 			}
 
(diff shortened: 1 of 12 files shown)


rivo:
Hi Tony,

I appreciate your initiative here. But I'm afraid I'm not going to accept this pull request, at least not as it is at the moment. Generally speaking, it introduces complexity to the package where I don't see the benefit. (Unfortunately, you didn't elaborate in the comments on the benefits you see so it's difficult to judge.) It redefines the semantics of functions (e.g. `SetInputCapture()`), leaving their definition in an ambiguous state. And finally, the added code doesn't follow the structure of the existing code, i.e. functions with abbreviated names (e.g. `DelItem()`) and functions without any comment (and the comments of the functions you changed have not been adapted) or with insufficient comments.

If you want to get into individual topics, I'm happy to discuss them with you. I would suggest that you open individual issues for any of the following items which are important to you:

- Use cases for needing direct access to `FlexItems`. I'm not sure I understand why this is needed. An example where you run into limitations with the current implementation would be great.
- Same with `ListItems`. Please provide an example.
- Reasons for replacing `EventKey` with `Event`. What specific events do you need access to that you don't at the moment?
- "Mounting": What essential functionality does this add? What is a "router"? How is this needed?
- Why do all primitives need a property map? How is this used? Can you provide examples where this is needed?
- What do we need a primitive ID for?

Cheers!

verdverm:
Some usage is here: https://github.com/verdverm/vermui

I mostly intended to start a discussion around these things. (Why comments/demos/test are MIA) I have a more advanced use case coming out soon. I will put up an asciinema sooner, some use cases in the mean-time. I've gotten into a habit of creating an experimental branch and then chipping off smaller contributions and PRs.

EventKey -> Event, the EventMouse and custom events

FlexItems/Life-cycle, I have hidden grid elements which can be shown with a hot-key. Some of these hidden elements are connected to data streams and continually update in the background. Others have initialized a data stream source, and then turn it on/off when when the show/hide.

A lot of this is Reactish inspired. Hence the additions for life-cycle, props, and unique IDs which will be helpful to getting things working and optimzed.

PR #51, opened 2018-02-06
added keyHandlers for overriding specific keys on input fields

pjtatlow:
I need the ability to intercept specific keys on an input field to do specific things. and this is my solution! It allows you to register a function with the input field as a handler for any given key, and in the input field's InputHandler function it runs that function. Handler functions are able return true or false to tell the InputHandler to continue handling that key event as normal or not.


rivo:
This already exists: https://godoc.org/github.com/rivo/tview#Box.SetInputCapture

Since all primitives inherit from `Box`, you can call this function on all of them, including `InputField`.

Let me know if this helps.

PR #53, opened 2018-02-06
vendor dependencies with dep

jmank88:
This PR proposes vendoring dependencies with [`dep`](https://github.com/golang/dep). `Gopkg.toml` was generated by `dep init`, then I removed the example comments and added `non-go` pruning.


jmank88:
> even if there was a good reason to include Gopkg.toml, it would be better if users could generate their own vendor directory from it. (I'm not familiar enough with dep to know if that can be done.)

It can. The normal workflow is to run `dep ensure` to generate the lock and vendor directory, but the `-no-vendor` and `-vendor-only` options can be used to split the workflow so that only the lock file is generated, and users can just run `dep ensure -vendor-only`.  I can adjust the PR if you'd like.

rivo:
Thanks for the explanation. That's good to know.

Regarding the PR itself, I'd still like to know if we're solving an actual problem here. Adding dependency information will force me to (A) install and use `dep` and (B) maintain the `dep` configurations in the long run (assuming that you won't do this for this project). I would certainly take on this additional work if I knew that a significant number of users needed it. So far, I haven't seen any actual requests for this so maybe you can shed some light as to why we need this.

freman:
I am of the opinion that for the most part the dependancy management in go with `go get` is sufficient for a maintainer of a package if they're keeping it up to date with the tip of other packages.

Chances are if someone wants to lock their dependancies to a version with something like `dep` then they're going to be locking this package too as part of their project requirements.

(personally, I haven't found a usecase for package management or vendoring, and where I've worked with vendoring in the past it's been a `windows xp` nightmare - no-one wants to update x because of y because of z, which hasn't been updated since the dawn of time and no-one has the time to update. Without vendoring z would have been updated every time it caused a problem.)

rivo:
I tend to agree here so, given the lack of a counter-argument, I'll close this for now. If someone provides a good reason to add vendoring to this repo, I'm happy to reopen this pull request.

Issue #58, opened 2018-02-18
Question: How do I arbitrarily position primitives and leverage tview.Application? 


zaquestion:
How do you draw multiple boxes at arbitrary coordinates? 

I'm trying to create something like this. The below I did with https://github.com/jroimartin/gocui which makes it pretty easy to just draw some boxes on the screen, but working with primitives is lacking. 
![pipeline](https://user-images.githubusercontent.com/2358914/36355540-af2fd158-1499-11e8-8d0f-9ed86f5765a1.png)

tview seems to have the right ideas for working with primitives, but I can't figure out how to draw multiple boxes at arbitrary coordinates. I saw @rivo talking about laying out primitives by yourself in https://github.com/rivo/tview/issues/57#issuecomment-366508543 , but I've got some disconnect between starting the application and drawing the boxes.

I don't really have any meanful code snippets, I've just been going through different iterations
```
	root := tview.NewBox().SetBorder(true)
	if err := tview.NewApplication().SetRoot(root, true).Run(); err != nil {
		log.Fatal(err)
	}

        // try to mac more boxes
	b := tview.NewBox().SetBorder(true)
	b.SetRect(20, 20, 20, 20)
```
Also tried creating my own screen, but without the event loop the program exits immediately
```
	s, err := tcell.NewScreen()
	if err != nil {
		log.Fatal(err)
	}
	b := tview.NewBox().SetBorder(true)
	b.SetRect(20, 20, 20, 20)
	b.Draw(s)
```

zaquestion:
Okay might have given in too soon. Just saw https://github.com/rivo/tview/issues/47 which suggested using `Page`s and that seems to let me draw boxes with arbitrary size/position. Still haven't broken into drawing lines, but I suspect I'll be able to call into `tcell` for that. If this sounds like the right approach feel free to close this out.

rivo:
To give some background, the `Application` takes one root primitive. But you obviously have more than one primitive so you'll need to use one of the container primitives such as `Flex` or `Pages` (and soon `Grid`). `Flex` overrides the position of the contained primitives so this is not what you want. That leaves you with `Pages` (with the `resize` flag set to `false`).

Drawing lines is probably a bit more difficult. You could use the `TextView` or create your own `Primitive`. The latter will give you direct access to the `Screen` object but is a bit more sophisticated. I meant to write a Wiki entry on how to write your own `Primitive` but haven't gotten around to it yet.

[...]

PR #59, opened 2018-02-20
(application) expose tcell.Screen through getter

zaquestion:
This is largely for reference and isn't expected to be merged as is.

Curious what your thoughts are on exposing the screen to the user? Currently its difficult to do so since `Run()` creates the screen and starts the event loop. As a proof of concept I moved it into `NewApplication`, but needed add an error to the return. This kind of API breakage probably isn't the best idea, so perhaps this functionality would be best served in a seperate function (like `NewApplicationScreen()` and `Run()` updated to check for a screen before creatinga a new one.

### DIFF
--- a/application.go
+++ b/application.go
@@ -33,8 +33,26 @@ type Application struct {
 }
 
 // NewApplication creates and returns a new application.
-func NewApplication() *Application {
-	return &Application{}
+func NewApplication() (*Application, error) {
+	var (
+		a   Application
+		err error
+	)
+
+	// Make a screen.
+	a.screen, err = tcell.NewScreen()
+	if err != nil {
+		return nil, err
+	}
+	if err := a.screen.Init(); err != nil {
+		return nil, err
+	}
+
+	return &a, nil
+}
+
+func (a *Application) Screen() tcell.Screen {
+	return a.screen
 }
 
 // SetInputCapture sets a function which captures all key events before they are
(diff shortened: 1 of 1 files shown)


rivo:
Please have a look at the new `SetBeforeDrawFunc()` and `SetAfterDrawFunc()` functions in `Application`. They give you direct access to the screen every time the root primitive is drawn.

I guess my main issue with your suggestion (apart from the breaking change) was that it would be difficult to synchronize screen updates if you keep a reference to the screen in parallel to everything else that is happening in `Application`. The new callbacks should solve that.

Regarding line drawing (#58), there is a new function [`PrintJoinedBorder()`](https://godoc.org/github.com/rivo/tview#PrintJoinedBorder) which could help you. I needed this for the `Grid` class so it's suitable mostly for drawing overlapping rectangles, and it currently only handles single, thin lines. But it could make it easier for you to achieve your goal. Eventually, I want to extend it to handle other line types, too. (It's a lot of combinations and if there's a system to the unicode code points, I haven't found out yet.)

zaquestion:
@rivo Just wanted to shoot you a quick follow up, `SetAfterDrawFunc` worked perfectly for this case. I was able to replicate exactly what I had before with minimal refactors from what the pull request exposed. The end result was also cleaner code :)

I'm gonna try out `PrintJoinedBorder()` next to refactor the lines, I'll respond in #58 on how that goes. Thanks for being so responsive and providing a way for my usecases to work!

rivo:
Glad you were able to solve your problem.

Yes, you'll need to clear the screen like that. There's a [comment](https://godoc.org/github.com/rivo/tview#Application.SetBeforeDrawFunc) that clarifies this.

PR #62, opened 2018-02-28
Expandable column

dmke:
I've added a bit of code to mark a column as "expandable", i.e. to use the remaining horizontal space, effectively making any table a full-width table.

My use case is [`multiping`](https://github.com/digineo/go-ping/tree/master/cmd/multiping), which should align the statistic column on the right, and the host information on the left (kind of like [`mtr`](https://github.com/traviscross/mtr))

expanded

For now, only one column can be made "expandable", but the API could later be upgraded to `func (t *Table) SetExpandable(columns ...int) *Table` to allow sharing the remaining space between multiple columns.

### DIFF
--- a/demos/presentation/table.go
+++ b/demos/presentation/table.go
@@ -116,6 +116,38 @@ const tableSeparator = `[green]func[white] [yellow]main[white]() {
         [yellow]Run[white]()
 }`
 
+const tableExpanded = `[green]func[white] [yellow]main[white]() {
+    table := tview.[yellow]NewTable[white]().
+        [yellow]SetFixed[white]([red]1[white], [red]1[white]).
+        [yellow]SetExpanded[white]([red]3[white])
+    [yellow]for[white] row := [red]0[white]; row < [red]40[white]; row++ {
+        [yellow]for[white] column := [red]0[white]; column < [red]7[white]; column++ {
+            color := tcell.ColorWhite
+            [yellow]if[white] row == [red]0[white] {
+                color = tcell.ColorYellow
+            } [yellow]else[white] [yellow]if[white] column == [red]0[white] {
+                color = tcell.ColorDarkCyan
+            }
+            align := tview.AlignLeft
+            [yellow]if[white] row == [red]0[white] {
+                align = tview.AlignCenter
+            } [yellow]else[white] [yellow]if[white] column == [red]0[white] || column >= [red]4[white] {
+                align = tview.AlignRight
+            }
+            table.[yellow]SetCell[white](row,
+                column,
+                &tview.TableCell{
+                    Text:  [red]"..."[white],
+                    Color: color,
+                    Align: align,
+                })
+        }
+    }
+    tview.[yellow]NewApplication[white]().
+        [yellow]SetRoot[white](table, true).
+        [yellow]Run[white]()
+}`
+
 const tableBorders = `[green]func[white] [yellow]main[white]() {
     table := tview.[yellow]NewTable[white]().
         [yellow]SetFixed[white]([red]1[white], [red]1[white]).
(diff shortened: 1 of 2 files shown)


rivo:
Thanks. I understand the request. It's similar to what is already available with the `Grid` class. Let me think about the best way to solve this.

I'd rather keep this in line with similar features in `Grid` and `Flex` where you can set a proportional width for multiple columns. Also, all column-based attributes in `Table` are derived from the `TableCell` class and I wouldn't want to make an exception for this case. (You suggested upgrading the `SetExpandable()` function in the future but I'm trying to avoid breaking changes so a good solution should be found now already.)

Give me some time and I'll figure out a solution to this.

dmke:
> I'd rather keep this in line with similar features in Grid and Flex where you can set a proportional width for multiple columns.

Understandable, I too wanted to keep the API similar to existing functionality, but I also wanted it to be *simple* and *unintrusive*. I'm open for suggestions though, and willing to reimplement this with another API.

> You suggested upgrading the SetExpandable() function in the future but I'm trying to avoid breaking changes so a good solution should be found now already.

Yeah, my suggested upgrade would be backward-compatible. The API of this PR is

```go
func (t *Table) SetExpandable(int) *Table
```

and the upgrade would be

```go
func (t *Table) SetExpandable(...int) *Table
```

In both cases, the consumers would call

```go
table := tview.NewTable().SetExpandable(1)
```

to mark the column with index 1 to be expandable.

The "upgrade" would then also allow the consumer to mark multiple columns expandable:

```go
table := tview.NewTable().SetExpandable(1, 4)
```

I haven't come around implementing this yet, because the calculations for the remaining space are a bit fiddly, and easily introduce off-by-one errors, resulting in tables a character too wide or to narrow...

I'm also not sure how to handle cases where the remaining horizontal space is, say 3 characters, but you have more than 3 columns marked as expandable. The naïve solution might be to distribute an extra space to the first three columns—but that might not produce the visually most pleasing result. Imagine the first columns to be sparsely populated, i.e. they already have a "natural" spacing, while the other more densely populated columns would benefit more from extra whitespace... 

> Also, all column-based attributes in Table are derived from the TableCell class and I wouldn't want to make an exception for this case.

Hm. I don't know if expandability is a concern of a single `TableCell`. When moved to `TableCell`, the consumer would need to tell each cell to be expandable or not (this also introduces problems when cousumers add expandable cell in different columns).

Let me rephrase my goal: I need a way to tell the `Table` to use all of the horizontally available space. Then maybe the API could be something like this:

```go
func (t *Table) SetFullWidth(enabled bool, columnWeights ...int) *Table
```

where `enabled` toggles the behaviour, and `columnWeights` define the `Grid`/`Flex`-like proportional widths? Expanding column 3 in a 5 column-table then would be something like

```go
t.SetFullWidth(true, 1, 1, 100, 1, 1) // col 3 is at most 100 times as wide as cols 1,2,4,5
```

> Give me some time and I'll figure out a solution to this.

Sure, no problem :-)

rivo:
The idea of putting it in `TableCell` is to have a value, e.g. `Expansion` which is `0` by default. The column value is the maximum over all cells in that column (so you really only need to define it once, e.g. in the table header). The value will be used as a proportional weight indicating how much horizontal space will be added to the column if the table is smaller than the available width.

Other values in `TableCell` (e.g. `MaxWidth`) already work that way, i.e. the maximum is used for the column width, therefore I'd like this to be in line with that.

Also, if I only want to expand column `8`, I don't have to do this: `SetFullWidth(true,0,0,0,0,0,0,0,1)`.

dmke:
Ah, I see, thanks for the explanation.

Let me try to implement this.

rivo:
Actually, I was planning on implementing this myself anyway. Will probably have it ready for you by Monday.


Sent from my iPhone

rivo:
Please have a look at the referenced commit. This should be in line with what you requested.

I actually also tried implementing a "contraction" value which shrinks columns if the table is larger than the available width. But then I realized this is in conflict with the table's behaviour of moving columns off-screen and letting you navigate to them. So there was no easy solution to that. I guess using `Grid` would be better to achieve a width adaption in both directions.

dmke:
At a first glance, this looks good, thank you very much.

I'll try this out later when I'm back home.

dmke:
Works like a charm :-)

multiping

Issue #94, opened 2018-03-30
Feature Request: Improve method of retrieving form input


joegrasse:
It would be a lot easier to work with forms if you could get the input by name.

For example

```Go
firstname := form.GetFormItemByName("firstname").(*tview.InputField).GetText()
```

rivo:
I don't know. I guess. But the way it is now, I'd either have to duplicate a whole bunch of `Form` methods or break backwards compatibility to achieve this (i.e. by adding a name parameter to all `Add` methods).

I would suggest that you do this:

```go
firstName := tview.NewInputField()
form.AddFormItem(firstName)
```

Then you can always reference it by its actual variable `firstName` and you don't even have to cast it:

```go
name := firstName.GetText()
```

If it's just about getting the input field's text, I usually use the event callback to update its string variable:

```go
form.AddInputField("First name", "", 30, nil, func(text string) {
	name = text
})
```

Would this help you?

joegrasse:
Yes, that is what I do now in the ```form.AddButton("Save", func() {``` section. This makes it hard to break the code out that performs the save functionality though. I would have to pass all of the form items to the function, or do ```firstname := form.GetFormItem(0).(*tview.InputField).GetText()```. As you can see, 0 mapping to firstname isn't quite obvious.

Also, I guess what I really meant was ```firstname := form.GetFormItemByLabel("firstname").(*tview.InputField).GetText()```. This wouldn't break backwards compatibility, since the Add* methods have a label.

rivo:
Yeah, that's what I was going to offer you. I hope I'll have this for you tomorrow at the latest.

joegrasse:
Awesome. Thank you.

rivo:
Have a look, it's there now.

joegrasse:
Thanks again.

joegrasse:
@rivo I have been testing out your changes and have ran into an issue. After the form has been drawn, the label can change. As-in, it can have spaces added to it to line up all of the input boxes. [Here](https://github.com/rivo/tview/blob/cc84c982fc1fc91616feb488f80725fdcb76c266/form.go#L323) you are updating the item attributes and [above](https://github.com/rivo/tview/blob/cc84c982fc1fc91616feb488f80725fdcb76c266/form.go#L308) you are adding spaces to the label if needed. So the label doesn't always match what you set it to, which make ```GetFormItemByLabel``` hard to use at the moment.

joegrasse:
[...]

PR #99, opened 2018-04-09
Fix for GetFormItemByLabel Issue

joegrasse:
Fix for issue I encountered on https://github.com/rivo/tview/issues/94.

### DIFF
--- a/checkbox.go
+++ b/checkbox.go
@@ -1,6 +1,8 @@
 package tview
 
 import (
+	"strings"
+
 	"github.com/gdamore/tcell"
 )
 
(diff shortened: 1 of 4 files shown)


rivo:
I opted for a slightly different solution here. Instead of adding padding to the label, one can set a label width. The label text does not get modified anymore. This should make the `GetFormItemByLabel()` work.

joegrasse:
I just thought I would note that your change does cause a backwards incompatibility.

rivo:
Where does it? I think I'm not seeing it...

rivo:
`SetFormAttribrutes()` is meant to be used only by the `Form` class. I guess I could have made it private in the first place.

joegrasse:
Your latest changes work for me. Thank you.

PR #106, opened 2018-04-19
Documentation fix for flex.AddItem()

frankbraun:
The proportion must be at least 1 if fixedSize == 0.

P.S.: Very cool library, thanks a lot for writing it!


rivo:
Yes, good catch. Thank you.

PR #112, opened 2018-05-09
Fix panic with <Enter> on empty list

hoop33:
This PR fixes a panic (index out of range) if the focus is on an empty list and user hits <Enter>.


rivo:
Thanks!

Issue #101, opened 2018-04-17
missing character in thai


prasertwi:
i test  textview  code from wiki but  can't display correct thai character   
![tview](https://user-images.githubusercontent.com/1924296/38845962-246b6682-4224-11e8-9fb0-40bc45b1f1f4.png)   

and then i test with tcell code is  correct

![tcell](https://user-images.githubusercontent.com/1924296/38845985-3c5220ba-4224-11e8-99d3-f30c74b85786.png)

how to coding for solve problem ? Please suggestion for me 
thank you .

**tview code**
`package main

import (
	"fmt"
	"strconv"
	"strings"
	"time"

	"github.com/gdamore/tcell"
	"github.com/rivo/tview"
)

const corporate = `เป็นมนุษย์ สุดประเสริฐ เลิศคุณค่า
กว่าบรรดา ฝูงสัตว์ เดรัจฉาน
จงฝ่าฟัน พัฒนา วิชาการ
อย่าล้างผลาญ ฤๅเข่นฆ่า บีฑาใคร
ไม่ถือโทษ โกรธแช่งซัด ฮึดฮัดด่า
หัดอภัยเหมือน กีฬา อัชฌาสัย
ปฏิบัติ ประพฤติกฎ กำหนดใจ
พูดจาให้จ๊ะ ๆ จ๋า ๆ น่าฟังเอยฯ`

func main() {
	app := tview.NewApplication()
	textView := tview.NewTextView().
		SetDynamicColors(true).
		SetRegions(true).
		SetChangedFunc(func() {
			app.Draw()
		})
	numSelections := 0
	go func() {
		for _, word := range strings.Split(corporate, " ") {
			if word == "the" {
				word = "[red]the[white]"
			}
			if word == "to" {
				word = fmt.Sprintf(`["%d"]to[""]`, numSelections)
				numSelections++
			}
			fmt.Fprintf(textView, "%s ", word)
			time.Sleep(200 * time.Millisecond)
		}
	}()
	textView.SetDoneFunc(func(key tcell.Key) {
		currentSelection := textView.GetHighlights()
		if key == tcell.KeyEnter {
			if len(currentSelection) > 0 {
				textView.Highlight()
			} else {
				textView.Highlight("0").ScrollToHighlight()
			}
		} else if len(currentSelection) > 0 {
			index, _ := strconv.Atoi(currentSelection[0])
			if key == tcell.KeyTab {
				index = (index + 1) % numSelections
[...]

PR #117, opened 2018-05-16
a bug fix and add demo displaying all color as handy util for tview

stephencheng:
As the title suggested, it's a handy util to display all colors from tcell for developers to have a quick pickup

Hopefully this is useful to others


stephencheng:
@rivo adding fix for #101

rivo:
Please see #101 regarding the bug you found.

With respect to this demo, I don't think this should be in the repo. Have a look at what this looks like on my terminal:

image

This doesn't make it very useful to me. (I can't even read what's written in the white boxes.) Maybe it's better for someone whose terminal can display all colours.

But I'd be ok with linking to a Gist of yours from the `tview` Wiki given you make the following changes:

- Choose a contrasting colour for the cell texts (you can use the HCL colourspace for that, see [here](https://github.com/rivo/tview/blob/a26df4e781481707856db8904c8ff902c67aae5d/table.go#L819) for an example of how this is used). Alternatively, set the text to the desired color and leave the background black.
- Make the code pass all `go vet` tests. Especially your function names and some variable names don't follow the Go conventions.
- There's dead/uncommented code which should be removed.
- To save space, you'll probably want to call `SetBorders(false)` on the table. I would also remove the "code:" string in every cell as it's not needed and only uses up valuable space.

PR #126, opened 2018-05-26
Refactored semigraphics and configurable borders

benweidig:
Fixing horizontal/vertical in box.go, added comments to border runes for better visualization

### DIFF
--- a/box.go
+++ b/box.go
@@ -243,20 +243,20 @@ func (b *Box) Draw(screen tcell.Screen) {
 			bottomLeft = GraphicsDbBottomLeftCorner
 			bottomRight = GraphicsDbBottomRightCorner
 		} else {
-			vertical = GraphicsHoriBar
-			horizontal = GraphicsVertBar
+			vertical = GraphicsVertBar
+			horizontal = GraphicsHoriBar
 			topLeft = GraphicsTopLeftCorner
 			topRight = GraphicsTopRightCorner
 			bottomLeft = GraphicsBottomLeftCorner
 			bottomRight = GraphicsBottomRightCorner
 		}
 		for x := b.x + 1; x < b.x+b.width-1; x++ {
-			screen.SetContent(x, b.y, vertical, nil, border)
-			screen.SetContent(x, b.y+b.height-1, vertical, nil, border)
+			screen.SetContent(x, b.y, horizontal, nil, border)
+			screen.SetContent(x, b.y+b.height-1, horizontal, nil, border)
 		}
 		for y := b.y + 1; y < b.y+b.height-1; y++ {
-			screen.SetContent(b.x, y, horizontal, nil, border)
-			screen.SetContent(b.x+b.width-1, y, horizontal, nil, border)
+			screen.SetContent(b.x, y, vertical, nil, border)
+			screen.SetContent(b.x+b.width-1, y, vertical, nil, border)
 		}
 		screen.SetContent(b.x, b.y, topLeft, nil, border)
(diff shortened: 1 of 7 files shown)


rivo:
Thanks. Please give me some time to review. I will get back to you soon.

benweidig:
No hurry. Starting to use the changes in my own app so I can see if it works as intended, the demos looked good so far.

box.go
```diff
@@ -236,27 +236,27 @@ func (b *Box) Draw(screen tcell.Screen) {
...
+			horizontal = Borders.HorizontalFocus
+			vertical = Borders.VerticalFocus
+			topLeft = Borders.TopLeftFocus
+			topRight = Borders.TopRightFocus
+			bottomLeft = Borders.BottomLeftFocus
+			bottomRight = Borders.BottomRightFocus
 		} else {
-			vertical = GraphicsHoriBar
-			horizontal = GraphicsVertBar
-			topLeft = GraphicsTopLeftCorner
```
rivo: I never noticed that I mixed up vertical and horizontal here. Thanks for catching this.

semigraphics.go
```diff
@@ -0,0 +1,234 @@
...
+	BoxDrawingsLightLeftAndHeavyRight             rune = '\u257c' // ╼
+	BoxDrawingsLightUpAndHeavyDown                rune = '\u257d' // ╽
+	BoxDrawingsHeavyLeftAndLightRight             rune = '\u257e' // ╾
+	BoxDrawingsHeavyUpAndLightDown                rune = '\u257f' // ╿
+)
+
+// SemigraphicJoints is a map for joining semigraphic (or otherwise) runes.
+// So far only light lines are supported but if you want to change the border
+// styling you need to provide the joints, too.
+var SemigraphicJoints = map[string]rune{
```
rivo: Have you tested all of these joints? Or did you generate them from the previous joint list? I want to be sure that these combinations are correct.

semigraphics.go
```diff
@@ -0,0 +1,234 @@
...
+	if result == 0 {
+		result = ch
+	}
+
+	// We only print something if we have something.
+	screen.SetContent(x, y, result, nil, style)
+}
+
+// PrintJoinedBorder is the backwards-compatible wrapper for the new PrintJoinedSemigraphics.
+func PrintJoinedBorder(screen tcell.Screen, x, y int, ch rune, color tcell.Color) {
```
rivo: I think this can be removed. I know this will break compatibility but it wasn't a core function of the package anyway. (I'd be surprised if anybody uses it.)

util.go
```diff
@@ -19,90 +19,29 @@ const (
 	AlignRight
 )
 
-// Semigraphical runes.
+// Deprecated semigraphical runes.
+// You should use the runes provided by semigraphical.go or Borders directly.
 const (
```
rivo: Let's remove these, too.

semigraphics.go
benweidig: I "search&replace"-d them, and the table demo looked good. But I will check again and maybe add comments with the cimbined runes for a better understanding.

benweidig:
Thx for the feedback, I'll change/update the PR accordingly in the next few days.

benweidig:
- SemigraphicJoints: Grid and Table demo are looking good, the comments were created with "Jump to Definition", so the Joints should all be correct. Also I've added a comment about the order of the runes, I haven't realized it before that it's important.

- Removed the PrintJoinedBorder and changed Grid accordingly

rivo:
Looks good. I merged it. Thanks again for the effort.

PR #135, opened 2018-06-20
Block main event loop with mutex during Suspend() call

### LINKED ISSUE #134: Bug: Suspend() doesn't work from goroutine
I'm trying to get user input using `fmt.Scanf()` by temporarily suspending tview/tcell application and later resuming it. If `Suspend()` is called from same thread with `Run()`/`KeyboardInterceptor()` - everything works, I can get user input, but application hangs (`Run()` finishes) when goroutine is used. Following sample code demonstrates the issue:

```golang
package main

import (
	"fmt"
	"log"
	"os"

	"github.com/gdamore/tcell"
	"github.com/rivo/tview"
)

func doInput(app *tview.Application) {
	app.Suspend(func() {
		fmt.Printf("Input something: ")
		var code string
		if _, err := fmt.Scan(&code); err != nil {
			log.Fatalf("Unable to read authorization code %v", err)
		}
		fmt.Printf("\r\nInput: %s\r\n", code)
	})
}

func main() {
	os.Setenv("TERM", "xterm-256color")
	box := tview.NewBox().SetBorder(true).SetTitle("Hello, world!")
	app := tview.NewApplication()
	pages := tview.NewPages()
	pages.AddPage("grid", box, true, true)
	app.SetInputCapture(func(event *tcell.EventKey) *tcell.EventKey {
		switch event.Key() {
		case tcell.KeyCtrlR:
			doInput(app)
		default:
			return event
		}
		return event
	})
	if err := app.SetRoot(pages, true).Run(); err != nil {
		panic(err)
	}
}
```
If I call `doInput(app)` - everything is OK, but when I change it to `go doInput(app)` - application hangs

kvj:
Fixes #134 

Please take a look on it.
This replaces `suspended bool` flag with dedicated mutex, which simplifies code and fixes above mentioned bug

### DIFF
--- a/application.go
+++ b/application.go
@@ -41,7 +41,7 @@ type Application struct {
 	// was drawn.
 	afterDraw func(screen tcell.Screen)
 
-	// Used for events loop lock during Suspend()
+	// Halts the event loop during suspended mode
 	suspendMutex sync.Mutex
 }
 
@@ -183,13 +183,14 @@ func (a *Application) Suspend(f func()) bool {
 	a.Lock()
 
 	if a.screen == nil {
-		// Application is already suspended.
+		// Screen has not yet been initialized
 		a.Unlock()
 		return false
 	}
 
 	// Enter suspended mode.
 	a.suspendMutex.Lock()
+	defer a.suspendMutex.Unlock()
 	a.Unlock()
 	a.Stop()
 
@@ -209,12 +210,10 @@ func (a *Application) Suspend(f func()) bool {
 	var err error
 	a.screen, err = tcell.NewScreen()
 	if err != nil {
-		a.suspendMutex.Unlock()
 		a.Unlock()
 		panic(err)
 	}
 	if err = a.screen.Init(); err != nil {
-		a.suspendMutex.Unlock()
 		a.Unlock()
 		panic(err)
 	}
@@ -222,7 +221,6 @@ func (a *Application) Suspend(f func()) bool {
 	a.Draw()
 
 	// Continue application loop.
-	a.suspendMutex.Unlock()
 	return true
 }
 
--- a/application.go
+++ b/application.go
@@ -41,8 +41,8 @@ type Application struct {
 	// was drawn.
 	afterDraw func(screen tcell.Screen)
 
-	// If this value is true, the application has entered suspended mode.
-	suspended bool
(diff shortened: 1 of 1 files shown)


application.go
```diff
@@ -41,8 +41,8 @@ type Application struct {
 	// was drawn.
 	afterDraw func(screen tcell.Screen)
 
-	// If this value is true, the application has entered suspended mode.
-	suspended bool
+	// Used for events loop lock during Suspend()
```
rivo: Maybe a comment such as `Halts the loop during suspended mode .` is better here.

application.go
```diff
@@ -103,28 +103,20 @@ func (a *Application) Run() error {
...
-		}
 		a.Unlock()
 		if screen == nil {
+			a.suspendMutex.Unlock()
 			break
 		}
 
 		// Wait for next event.
 		event := a.screen.PollEvent()
+		a.suspendMutex.Unlock()
```
rivo: Doesn't it make sense to wrap the entire `for` block with the lock? It seems to me that the code after this `Unlock()` could also potentially interfere with suspended mode.

application.go
```diff
@@ -190,14 +182,14 @@ func (a *Application) Stop() {
 func (a *Application) Suspend(f func()) bool {
 	a.Lock()
 
-	if a.suspended || a.screen == nil {
+	if a.screen == nil {
 		// Application is already suspended.
```
rivo: This comment is now not correct anymore. It should probably read `Screen has not yet been initialized.`.

application.go
```diff
@@ -217,17 +209,20 @@ func (a *Application) Suspend(f func()) bool {
 	var err error
 	a.screen, err = tcell.NewScreen()
 	if err != nil {
+		a.suspendMutex.Unlock()
```
rivo: Instead of putting `Unlock()`'s everywhere, a `defer Unlock()` at the top is probably better.

application.go
kvj: No, in this case there will be a deadlock, if `Suspend()` will be called from `inputCapture()` handler (same goroutine)

kvj:
@rivo thanks for the feedback. I've updated PR

application.go
rivo: Right. That's true.

Issue #161, opened 2018-09-01
Incorrect rendering of Emojis with Zero-Width-Join characters


bcongdon:
It seems that some emojis sequences with zero-width-join runes are unable to be rendered correctly.

For example, with the List view:
```go
package main

import (
	"github.com/rivo/tview"
)

func main() {
	app := tview.NewApplication()
	list := tview.NewList().
		AddItem("👩‍🍳", "Some explanatory text", 'a', nil).
		AddItem("👨‍👨‍👧", "Some explanatory text", 'a', nil).
		AddItem("🏳️‍🌈", "Some explanatory text", 'a', nil).
		AddItem("Quit", "Press to exit", 'q', func() {
			app.Stop()
		})
	if err := app.SetRoot(list, true).Run(); err != nil {
		panic(err)
	}
}
```

These emojis are rendered as their component emojis, instead of the full emoji sequence.
![screen shot 2018-09-01 at 10 53 43 am](https://user-images.githubusercontent.com/706257/44948603-7fe2a800-add5-11e8-85a5-7a821e74b36e.png)

I've double checked that this is not a problem with my terminal emulator, as using the native `fmt.Println` function on these emojis renders them correctly.
```go
fmt.Println("Emojis: 👩‍🍳 👨‍👨‍👧 🏳️‍🌈")
```

![screen shot 2018-09-01 at 10 54 31 am](https://user-images.githubusercontent.com/706257/44948600-7bb68a80-add5-11e8-9fcd-ea82703917ea.png)

I've also noticed this behavior with the Table view, so I would suspect that this behavior is present in other views as well.

rivo:
Thanks for spotting this. Approximately 90% of my work on `tview` is spent on making Unicode characters work properly. I thought I had solved it but there are obviously still issues.

I'll look into this. Might take me some time if I find there's no simple fix.

rivo:
Looking into this some more, I found that I can't solve this without the help of [`tcell`](https://github.com/gdamore/tcell) and [`runewidth`](https://github.com/mattn/go-runewidth) (see referenced issues). Both of these packages seem to have issues with unicode sequences containing zero-width joiners.

Once support for ZWJs is implemented in these packages, I can fix it in `tview` as well. I hope that the maintainers will add support for ZWJ.

rivo:
I just added support for zero-width joiners in `tview`. Note that the width of these characters will still be wrong. Once @mattn fixes mattn/go-runewidth#20, this should go away.

[...]

Issue #103, opened 2018-04-17
Feature Request: Add ability to use the left and right arrow keys in inputFields


joegrasse:
It would be nicer if the user could use the left and right arrow keys within a form inputField to correct mistypes.

rivo:
I'm giving this the "on roadmap" label. I'm planning to add this.

Can't say exactly when but hopefully soon.

rivo:
The latest commit introduces free movement within input fields. This was an effort of many hours, much more than what I had anticipated. The reason is mostly Unicode support, in which one character can consist of multiple runes (with modifiers, zero-width joiners etc.) and one rune can occupy multiple bytes in a string. Additionally, the width of a character can be more than one cell on the screen. All of this needs to be handled properly.

To navigate the input field, I took cues from Bash:

- __Left arrow:__ Move left by one character.
- __Right arrow:__ Move right by one character.
- __Home, Ctrl-A, Alt-a:__ Move to the beginning of the line.
- __End, Ctrl-E, Alt-e:__ Move to the end of the line.
- __Alt-left, Alt-b:__ Move left by one word.
- __Alt-right, Alt-f:__ Move right by one word.
- __Backspace:__ Delete the character before the cursor.
- __Delete:__ Delete the character after the cursor.
- __Ctrl-K:__ Delete from the cursor to the end of the line.
- __Ctrl-W:__ Delete the last word before the cursor.
- __Ctrl-U:__ Delete the entire line.

If you find any undesirable behaviour, please open an issue.

joegrasse:
Thank you very much for your work on this!

PR #171, opened 2018-10-08
Allow left and right movement within inputfield

Asa-Dawson:
I've patched inputfield.go to allow left and right movement and editing the input's text from the cursor's position. This is my first patch for this codebase so I'm sure it's not up to scratch but I'm 100% happy to make changes and resubmit.

### DIFF
--- a/inputfield.go
+++ b/inputfield.go
@@ -288,8 +288,8 @@ func (i *InputField) getCursor() (int, int) {
 		rightLimit -= 2
 	}
 	buffer := 2 / (i.fieldWidth - 1)
-	rightText := i.cursor + buffer
-	leftText := int(math.Max(0, float64(i.cursor-buffer)))
+	rightText := clamp(i.cursor+buffer, 0, len(i.text)-1)
+	leftText := clamp(i.cursor-buffer, 0, len(i.text)-1)
 	fieldWidth := runewidth.StringWidth(i.text[leftText:rightText])
 	if i.fieldWidth > 0 && fieldWidth > i.fieldWidth-1 {
 		fieldWidth = i.fieldWidth - 1
--- a/inputfield.go
+++ b/inputfield.go
@@ -270,7 +270,7 @@ func (i *InputField) Draw(screen tcell.Screen) {
 		rightText := clamp(i.cursor+buffer, 0, len(i.text))
 		leftText := clamp(i.cursor-buffer, 0, len(i.text))
 		fieldText := i.text[leftText:rightText]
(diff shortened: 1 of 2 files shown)


rivo:
Thanks. I really appreciate your effort. Unfortunately, the changes this needs to make it mergeable are significant. Apart from a few small things, the main issue is that you can't assume the `InputField`'s text to be just one-cell-width characters represented by one byte. In other words, things like `len(i.text)` or `i.text[i.cursor:]` don't work with lots of Unicode characters which (A) will often consist of multiple bytes and even multiple runes and (B) don't always have a width of one screen cell. See #9, #101, and #161 for examples.

I'm guessing that making it work properly with Unicode characters will require you to rewrite almost the whole thing so I'm not sure it's worth it.

I was about to start to work on this myself now anyway. But if I do it, it would mean closing your pull request, i.e. not merging it. So I'd like to hear your thoughts on that first.

Asa-Dawson:
I've been dogfooding with this merge locally and I've recognised the issues you're talking about alongside a bug involving the field width causing wrapping bugs. 

Is there a consistent way to split up the input field's text into 1-cell-wide character collections? if so I'm happy to work on porting this from the naive one-byte approach if you're happy for me to continue doing that.

Otherwise if you have a better/preferred method or want to work on it yourself I'm totally happy for you to close this without merging, I'm not concerned about my contribution particularly I just want to make tview better :)

rivo:
Thanks. Again, I appreciate your contribution, even if I don't end up merging it. (Do not let this discourage you.)

Some characters are more than one cell wide. (Lots of Chinese characters are.) Some are modifiers which are 0 cells wide but must be combined with other characters. Then there are combinations of them (see "zero-width joiners"). It took me a whole bunch of issues to figure it all out. All of these must be considered.

As mentioned, I'm starting work on this now anyway. So I don't think you need to start a complete rewrite of your PR (which I have a feeling it would be).

I'm keeping this open until the whole thing is resolved. I might be able to give you more feedback in general once I've dug into this issue.

rivo:
My latest commit closes your PR. I hope this is ok for you. This was a major effort, much less simple than I thought. I've given some more details here: https://github.com/rivo/tview/issues/103#issuecomment-430894330

Thanks again for your contribution. Actually, your PR was what finally triggered me to tackle this issue.

Asa-Dawson:
Closing the PR is fine by me, I'm very happy to see tview improving! 🎈

Thanks for your hard work!

PR #163, opened 2018-09-07
Table thread safe fix

joegrasse:
Resolves #148

PR #172, opened 2018-10-09
make primitive updates thread-safe using channels

### LINKED ISSUE #148: panic: tview.(*Table).Draw.func1
I am getting a panic in tview. It looks like it might be in table.go line [469](https://github.com/rivo/tview/blob/6614b16d9037759065eac63d8fb3d8408e63e222/table.go#L469), but it is really hard to troubleshoot because of the way the panic is being printed.

panic

(Click image for bigger view)

3cb:
Resolves #148 as an alternative  to #163 

Happy to resubmit if I've overlooked something.

### DIFF
--- a/application.go
+++ b/application.go
@@ -46,11 +46,20 @@ type Application struct {
 
 	// Halts the event loop during suspended mode.
 	suspendMutex sync.Mutex
+
+	// Used to send screen events from separate goroutine to main event loop
+	events chan tcell.Event
+
+	// Used to send primitive updates from separate goroutines to the main event loop
+	updates chan func()
 }
 
 // NewApplication creates and returns a new application.
 func NewApplication() *Application {
-	return &Application{}
+	return &Application{
+		events:  make(chan tcell.Event, 100),
+		updates: make(chan func(), 100),
+	}
 }
 
 // SetInputCapture sets a function which captures all key events before they are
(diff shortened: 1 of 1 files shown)


rivo:
So I merged this but it required lots of modifications to make it work:

- `app.Stop()` did not properly stop the application because the event goroutine kept polling for events. Specifically, the following resulted in a messed up terminal (on macOS at least):

```go
func main() {
	box := tview.NewBox().SetBorder(true)
	app := tview.NewApplication()
	go func() {
		time.Sleep(3 * time.Second)
		app.Stop()
	}()
	if err := app.SetRoot(box, true).Run(); err != nil {
		panic(err)
	}
}
```

- Suspended mode also didn't work anymore. We need to be able to enter suspended mode from a goroutine or in response to a key event. Both posed different problems which needed to be solved.
- I removed `a.Draw()` after the execution of an updater function. I don't think it's desirable in all cases to refresh the screen after each update. Users should decide themselves whether the screen should be updated after each update or at a later time.
- I believe there could still have been a race condition if `app.Draw()` was called while an updater function executed. During the execution, there is no lock on `app` (and we don't want that) so `app.Draw()` could go ahead and read from primitives that would be modified at the same time. So I had to make a queue for `app.Draw()` as well.
- Similarly, some changes needed to be made to `TextView` as it implements `io.Writer` which could be potentially written to in a goroutine.
- Obviously, this required detailed documentation which I added to the package.

Nonetheless, thanks for your submission. The general structure you proposed is still there. I've run a number of tests but may not have caught everything. Have a go at it with your own application and let me know if you encounter any trouble.

joegrasse:
@rivo I have upgraded to this method for my app and now my app feels sluggish. The UI doesn't respond as quickly to my key presses. Should I open a new issue? Not sure what other information I could provide.

joegrasse:
@rivo Ok, I do have one other piece of info to add. If I remove the call to ```app.QueueUpdate``` (which would cause the concurrency problems again), the sluggish problem goes away.

rivo:
I don't know what your application does exactly so I'm not sure where this is coming from. Do you experience the same sluggishness in the demos included with `tview`?

I wrote a more detailed Wiki entry about handling concurrency: https://github.com/rivo/tview/wiki/Concurrency

In that entry, I explain that all handlers (except `TextView.SetChangedFunc()`) are invoked from the main goroutine and therefore need no syncing. Since you mentioned the lag in response to key presses, I have a feeling that you used `QueueUpdate()` where it wasn't needed. (Yes, it would lead to a lag because the main loop would turn at least one more round before running your function.)

`QueueUpdate()` is only needed when you have your own goroutine which will run outside of the main application loop. For example, if you're piping an external stream into your application or if your goroutine processes signals that are not generated by `tview`.

If that's not the case here, however, I'd need a way to reproduce your observations.

3cb:
It seems I missed quite a bit, sorry about that.  I've just taken a quick look and haven't had a chance to run it on my own but have a couple of questions:

1)  Would it be better to have `Application.Draw()` send to the `Application.updates` channel rather than a separate `redraw` channel?  Because of the way select blocks work isn't it possible that a queued redraw will be read from the `redraw` channel and run before the updater function was read and run, even if the redraw was sent later?
If we use the same channel then the order of updates and redraws from the same goroutine are guaranteed:
```go
func (a *Application) Draw() *Application {
	// We actually just queue this draw.
    f := func() {
        a.draw()
    }
    a.updates <- f
    return a
}
```

2)  Would you be interested in adding a second method `Application.QueueUpdateRedraw` (maybe needs a better name)? This would give users a little more flexibility and not require explicit calls to `Application.Draw()` if they want redraws with every update:
```go
func (a *Application) QueueUpdateRedraw(upd func()) *Application {
	f := func() {
		upd()
		a.draw()
	}
	a.updates <- f
	return a
}
```

joegrasse:
@rivo I found and fixed my problem. The function I was sending to the ```QueueUpdate()``` had a sleep for 1 second in it. I broke that out and now it works.

3cb:
@rivo updated today.  Everything seems to work well.

rivo:
@3cb Thanks for the suggestions. Of course, it's simpler to use the `update` queue for `Draw()` requests. Didn't think of that. I also added `QueueUpdateDraw()`, makes perfect sense.

PR #175, opened 2018-10-27
Do not duplicate dropdown option entry if it already exists

joakim-ribier:
Hi,

I propose this little fix (feature) for to not add a duplicate entry to a dropdown widget if the text already exists.

Sorry for advance, i don't know if this is the right way to proceed...

Thanks

### DIFF
--- a/dropdown.go
+++ b/dropdown.go
@@ -189,9 +189,24 @@ func (d *DropDown) GetFieldWidth() int {
 // AddOption adds a new selectable option to this drop-down. The "selected"
 // callback is called when this option was selected. It may be nil.
 func (d *DropDown) AddOption(text string, selected func()) *DropDown {
-	d.options = append(d.options, &dropDownOption{Text: text, Selected: selected})
-	d.list.AddItem(text, "", 0, nil)
+	// Do not duplicate option entry if it already exists
+	if d.notContains(text) {
+		d.options = append(d.options, &dropDownOption{Text: text, Selected: selected})
+		d.list.AddItem(text, "", 0, nil)
+	}
 	return d
+}
+
+// Check if option value already exists or not (case-insensitive)
+func (d *DropDown) notContains(text string) bool {
+	present := false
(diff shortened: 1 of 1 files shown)


rivo:
Hi, thank you for contributing to `tview`.

I have trouble understanding why you would like this functionality to be part of this package. I am not sure that users will expect this behaviour from the `DropDown` primitive.

Maybe you could provide an example where the current `tview` implementation was giving you headaches and we'll take it from there.

joakim-ribier:
Hi,

Actually, there is no directly a problem with the lib, i just thought it was more convenient to avoid entry duplicates directly in the primitive and not in my app side...

It's a bit heavy for each dropdown to check that it is not possible to add twice the same value :-)

rivo:
I can't think of a situation where one would accidentally add the same value twice. Can you give me an example where this happens for you?

joakim-ribier:
Hi,
Yes, of course i can :smiley:!

For the little context, to make my first app with the Go language, i try to impement my first text ui application with tview library.
 
It's a rich http client. So, my user needs to add header value in his request. For that, i implement a dropdown which list header values and an inputtext for adding a new value by the user.

So, if user adds twice the same value (for example my user clicks twice on "save" button on first screenshot), then, the dropdown contains now, several times the x value.

I think it's my own problem and not a library specifc solution for that...maybe the user may want a duplicate or not entry in the dropdown primitive :+1: 

You can close the PR :smiley: 

_form to add new value_
add-new-values

_example of dropdown values_
list-new-values

rivo:
I would actually agree with you here that this functionality should be part of your application rather than the `tview` package. You could have a set like this:

```go
set := make(map[string]struct{})
```

Then you also add your entry to this set: `set[option] = struct{}{}`. To test if an entry already exists, do this: `_, exists := set[option]`. If `exists == true`, you don't add your entry.

I'll close this PR now. If you have any more questions, feel free to post them here. I read them.

PR #177, opened 2018-10-30
Inputfield fix

### LINKED ISSUE #176: inputfield: unable to enter : (colon)
After the 990bc2ebcf815c84073a4fb08c08144808befa7f commit to allow navigation in input fileds, I can no longer enter a colon. I have traced it down to the code starting [here](https://github.com/rivo/tview/blob/990bc2ebcf815c84073a4fb08c08144808befa7f/inputfield.go#L365).

```
			modifiers := event.Modifiers()
			if modifiers == tcell.ModNone {
				ch := string(event.Rune())
				newText := i.text[:i.cursorPos] + ch + i.text[i.cursorPos:]
				if i.accept != nil {
					if !i.accept(newText, event.Rune()) {
						break
					}
				}
				i.text = newText
				i.cursorPos += len(ch)
			} else if modifiers&tcell.ModAlt > 0 {
				// We accept some Alt- key combinations.
				switch event.Rune() {
				case 'a': // Home.
					home()
				case 'e': // End.
					end()
				case 'b': // Move word left.
					moveWordLeft()
				case 'f': // Move word right.
					moveWordRight()
				}
			}
```

I changed it to the following and it works again.

```
			ch := string(event.Rune())
			newText := i.text[:i.cursorPos] + ch + i.text[i.cursorPos:]
			if i.accept != nil {
				if !i.accept(newText, event.Rune()) {
					break
				}
			}
			i.text = newText
			i.cursorPos += len(ch)
```

I suspect the : (colon) key has a modifier in windows.

joegrasse:
Fixes #176


rivo:
See https://github.com/rivo/tview/issues/176#issuecomment-435822656.

PR #186, opened 2018-11-18
fix #185

### LINKED ISSUE #185: Bug Fix - SetAcceptanceFunc prevents all key input
Running demos/inputfield/main.go won't allow any keyboard input. 

I traced the bug from demos/inputfield/main.go:16 SetAcceptanceFunc(tview.InputFieldInteger) 
to inputfield.go:367 where the following change worked for me:

```
367,368c367,368
< 			if i.accept != nil {
< 				return i.accept(newText, r)
---
> 			if i.accept != nil && !i.accept(newText, r) {
> 				return false
```

Thank you for such a beautifully designed and well written library!

delaneyj:
Fix per suggestion by @jr81


rivo:
Thanks!

PR #188, opened 2018-11-21
Box should have an content Primitive

delaneyj:
The Box primitive is used in many examples (especially `Flex`) but once you start using it adding contents inside the Box is not straight forward without generating custom `Primitives` and `.Draw()` functions.  This is an initial bare minimum implementation to support Box having a `.SetContents(p Primitive)`.  

This does not change how `b.draw` works but long term it may mean the need for a `DrawFunc` is unnecessary and allows for more composability.

### DIFF
--- a/box.go
+++ b/box.go
@@ -13,6 +13,9 @@ import (
 //
 // See https://github.com/rivo/tview/wiki/Box for an example.
 type Box struct {
+	// The contents to draw within the inner perimeter of the box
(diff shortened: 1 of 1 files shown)


delaneyj:
Ah, I see now.  Makes way more sense when you explain it like that.  Sorry for the confusion.

rivo:
No problem. Always happy to explain.

PR #190, opened 2018-11-22
Fixed example code in presentation demo textview

bcarlsson:
Example code would not run because app was undefined.


rivo:
Thanks!

PR #191, opened 2018-11-23
Add GetText() method

OriPekelman:
I find it useful to be able to get the nodes text in the handlers. Also seems more consistent with SetReference that has a GetReference.

### DIFF
--- a/treeview.go
+++ b/treeview.go
@@ -105,6 +105,11 @@ func (n *TreeNode) SetChildren(childNodes []*TreeNode) *TreeNode {
 	return n
 }
 
+// GetText returns this node's text.
+func (n *TreeNode) GetText() string {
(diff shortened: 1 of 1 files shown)


rivo:
Thank you for this. I'm generally ok merging this but I'd like to you ask you to make two changes:

- Please put the `GetText()` function underneath the `SetText()` function in the code. It's just a cosmetic change but I usually organize my code like this and it helps me (and others) navigate it better when related functions are together.
- `GetText()` should return a `string`, not an `interface{}`.

rivo:
No interest in this anymore?

OriPekelman:
Sorry it took me this long, I was on vacation. Is this better?

rivo:
Merged it. Thanks.

PR #196, opened 2018-11-26
Don't move cursor on inputfield.SetText() unless necessary

delaneyj:
The behavior seen was a separate go routine setting the text of a input field and that always moving the cursor to the end.  This changes leaves the cursor as is unless it most be bound to input limits.

### DIFF
--- a/inputfield.go
+++ b/inputfield.go
@@ -102,7 +102,12 @@ func NewInputField() *InputField {
 // SetText sets the current text of the input field.
 func (i *InputField) SetText(text string) *InputField {
 	i.text = text
-	i.cursorPos = len(text)
+
+	// If the new text is shorter than previous, set cursor to end
+	if tl := len(text); i.cursorPos > tl {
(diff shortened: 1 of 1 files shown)


rivo:
Thanks for your contribution. I have to say I'm not sure if this behaviour is what most users will expect. For example, if an input field is initially empty and some action populates it with text, I will want the cursor to automatically be at the end, not at the beginning of the field.

Maybe you had a different use case? Can you explain how in your application, the current behaviour was the wrong one?

delaneyj:
A perfect example is https://github.com/rivo/tview/pull/201.  If you move the cursor every time the arrow keys no longer work.

rivo:
Could you give me an example that doesn't depend on your other PR? As mentioned there, I had trouble running your demo code.

Maybe you can paste some short program here.

rivo:
Please open a new issue with the requested example if you would like to see this topic being followed up on.

PR #200, opened 2018-12-05
Add theme struct

delaneyj:
At beginning of application wanted to replace the `Styles` global and defining the struct made that clearer than initializing an anonymous struct.


rivo:
Thanks. The comments are a bit inconsistent but I'll fix that myself.

PR #201, opened 2018-12-05
Feature/checklist

delaneyj:
Add checklist


delaneyj:
Started work on a checklist.

rivo:
To be honest, I'm not sure that this is an essential primitive that needs to be part of `tview`. What you've implemented seems very specific to your own use case (is this actually a "todo list"?) and less adaptable to what others may need.

The code itself also doesn't follow the same structure/conventions used throughout this package and would need lots of adaptations, in my opinion. Even your demo code fails with a `runtime error: index out of range`.

If you find that you have trouble implementing your application in `tview`, I'd rather discuss how we can improve the `List` class (possibly along with `Form`) to enable what you need to do.

I should also note that you should be able to publish `tview` primitives in your own repo. It's why I made many of the internal functions and interfaces (e.g. `Primitive`, `Print()`, ...) public. So people can extend `tview` with primitives that have very specific functionality, even if I decide not to put them in the main repo.

rivo:
In the absence of any response to my comment, I'm closing this PR. You can always open a new issue to restart the discussion.

Issue #206, opened 2018-12-15
Pass option details into selection() callback when using (*DropDown).AddOption()


ardnew:
When using the `(*DropDown).SetOptions()` function, you can provide a `selected()` callback that is shared among all options because it passes in both the option text and index to the callback.

But if you use `(*DropDown).AddOption()`, the `selected()` callback is specific to that option, making it a little annoying when dynamically constructing the options.

I recommend also passing in both option text and index into the `selected()` callback when using `(*DropDown).AddOption()`. The convenience of this outweighs the risk of potentially having the wrong index passed into the callback if the user goes to the effort of manually moving options around after constructing the dropdown with `AddOption()` calls. Also it's consistent.

I think it's sufficient to change the following from:
```go
func (d *DropDown) AddOption(text string, selected func()) *DropDown {
	d.options = append(d.options, &dropDownOption{Text: text, Selected: selected})
	d.list.AddItem(text, "", 0, nil)
	return d
}
```

to something like (**warning**: untested):
```go
func (d *DropDown) AddOption(text string, selected func(string, int)) *DropDown {
	index := len(d.options)
	d.options = append(d.options, 
		&dropDownOption{
			Text: text, 
			Selected: func() {
				if selected != nil {
					selected(text, index)
				}
			}
		})
	d.list.AddItem(text, "", 0, nil)
	return d
}
```

And then update `(*DropDown).SetOptions()` to just pass it's `selected` argument directly into `(*DropDown).AddOption()` now that their prototypes are compatible.

rivo:
I appreciate you offering a solution already and I realize my response comes quite late. However, it's usually better to have a discussion first as I may want/need to come up with a different solution (requiring me to reject your PR).

In this specific case, as I understand it, you're looking for an item-independent "selected" callback which gives you the text and the item's index. Currently, you can only set this using `SetOptions()` but that doesn't work well when you want to modify options afterwards.

The callback provided with `AddOption()` is meant to allow you to provide specific functionality for *that one option*. The idea isn't for users to install the same callback for all options using `AddOption()`.

So instead of breaking backwards compatibility (and changing the general idea of this per-item callback), I suggest that I add a new function `SetSelectedFunc()` which installs a "global" handler for all options. It would receive the text and the option's index. Also, `SetOptions()` would set that handler instead of overriding all item's own handlers. And it would be in line with how `List` is implemented, for example.

Let me know if this helps you and I'll get to it then.

ardnew:
[...]

PR #208, opened 2018-12-18
change: provide drop-down option text and index as args to selection callback when using (*DropDown).AddOption()

ardnew:
rivo/tview#206

this change will break backwards compatibility for anyone using `(*DropDown).AddOption()` as its formal signature is changed. but i'm not convinced it's worth preserving the original and implementing this under a different name as the functionality is basically identical.

also I wasn't sure why `(*DropDown).SetOptions()` had the anon function in the body of its for-loop, so I took the liberty of removing it because it was a little confusing that variable identifiers were being reused at different scopes. 

but even if you want to restore it, one of the variables (`text` I believe) was being captured under the closure even though it was passed in via argument! so that needs to be changed regardless.

### DIFF
--- a/dropdown.go
+++ b/dropdown.go
@@ -186,10 +186,20 @@ func (d *DropDown) GetFieldWidth() int {
 	return fieldWidth
 }
 
-// AddOption adds a new selectable option to this drop-down. The "selected"
-// callback is called when this option was selected. It may be nil.
-func (d *DropDown) AddOption(text string, selected func()) *DropDown {
-	d.options = append(d.options, &dropDownOption{Text: text, Selected: selected})
+// AddOption adds a new selectable option to this drop-down and installs a
+// callback function which is called when this option was selected. It will be
+// called with the option's text and its index into the options slice. The
+// "selected" parameter may be nil.
+func (d *DropDown) AddOption(text string, selected func(string, int)) *DropDown {
+	index := len(d.options)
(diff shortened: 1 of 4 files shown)


ricochet1k:
This API seems very confusing to me. Why is there AddOption without any kind of ClearOptions() or RemoveOption()? Why does AddOption get its own selected func when SetOptions sets the same func for all? Why are you making a new func closure, when you could just pass the `currentOption` index and the option's text in when calling the Selected callback down on line 370ish?

ardnew:
i agree with most of your concerns -- obviously even submitting a PR for one of them.

> Why are you making a new func closure, when you could just pass the `currentOption` index and the option's text in when calling the Selected callback down on line 370ish?

that `Selected()` callback is defined in struct `listItem`, which is used throughout the library by multiple components (_not_ just this `DropDown` type). leaving out formal parameters in the callback effectively freezes the internal/private API for this reused type, so you don't have to worry about this function interface ever changing -- but you _will_ have to capture any associated data in a closure like he's doing (i'm not defending this design, just speculating).

i guess this philosophy isn't consistent though. if it were, this PR should instead remove the params from the callback in `SetOptions()` instead of adding them to `AddOption()`. but as a user, it would be pretty annoying to have to capture that data myself, so i would prefer to have them _**parameterized for the functions that get exported**_.

Issue #169, opened 2018-09-26
Adopt GO 1.11 Modules Versioning


mapl:
It would be nice to have some proper versioning for tview.

https://github.com/golang/go/wiki/Modules#new-concepts

rivo:
Support for modules is still preliminary and experimental in 1.11. Once it becomes official (most likely 1.12), I will very likely add support for it in this project (including versioning which, sigh, I suppose I won't be able to dodge). I'm waiting because things in the dependency management department have changed a lot (for a while, `go dep` was hailed as the future). So I'm waiting for the official and final word on this.

We've had a related discussion before, please see #116 and #53.

PR #211, opened 2018-12-21
Support Go Modules

SamWhited:
Hello,

Please consider supporting [Go Modules], the new packaging standard that will be adopted fully in Go 1.12. Experimental support is in Go 1.11 and the new module paths are supported in Go 1.9.7+ and Go 1.10.3+ in a read-only manner for backwards compatibility with all supported versions of Go.

Because this library is still below version 2 the [`go.mod`] file is fairly simple. The only other thing that would need to be done if this PR were merged is to create a semver compatible tag (eg. `v0.0.1`) that users of this library can pin to.

Thank you for your consideration.

[Go Modules]: https://github.com/golang/go/wiki/Modules
[`go.mod`]: https://tip.golang.org/cmd/go/#hdr-The_go_mod_file

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -1,0 +1,9 @@
+module github.com/rivo/tview
+
+go 1.9
+
+require (
+	github.com/gdamore/tcell v1.1.1
+	github.com/lucasb-eyer/go-colorful v0.0.0-20181028223441-12d3b2882a08
+	github.com/mattn/go-runewidth v0.0.4
+)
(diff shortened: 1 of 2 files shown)


rivo:
Please see https://github.com/rivo/tview/issues/169#issuecomment-425495341 and the linked issues for more infos.

SamWhited:
The release branch has been cut and a beta is out; this will not change before 1.12. Not to mention that it has been merged and released into all supported Go versions. Now is the time we're trying to get adoption up before it becomes the default in 1.12.

Issue #217, opened 2019-01-03
Do not scroll List items unless selected item is out of view


ardnew:
When navigating a `List` with more items than can fit on screen, the `Draw` routine tries to simulate a scrolling effect. However, this scrolling only works as expected in one direction -- when you scroll down.

When the user scrolls the list downward, the bottom-most item is highlighted/selected as the prior items scroll upward out of view _as expected_. 

When the user scrolls the list upward, the bottom-most item is _still_ highlighted/selected as the prior items scroll downward out of view, which is not what I think most people would expect.

For comparison, consider any given text editor. If your cursor is on the bottom-most row of the editor view, and you press the `Up` key on your keyboard, the cursor moves upward and the items visible on screen remain stationary until the cursor reaches the top of the screen -- at which point the whole relationship should happen in reverse.

Basically, unless the user's currently selected item is outside the range of visible items, the view should not be scrolling at all.

rivo:
See also https://github.com/rivo/tview/pull/218#issuecomment-456959828.

PR #218, opened 2019-01-03
Do not scroll List items unless selected item is out of view

ardnew:
This change adds a single new field `viewOffset` to the `List` struct because it is impossible (I'm _pretty_ sure) to determine which items should be drawn at any given instant with only the state information available when `Draw()` is called.

This new `viewOffset` field effectively remembers where the view existed the _last_ time `Draw()` was called, and from that along with the currently selected item, determines which items need to be drawn.

Please test this out, it felt very natural to me, but I haven't looked to see if `Table` or `TreeView` would also benefit from the same logic.

This resolves issue #217

### DIFF
--- a/list.go
+++ b/list.go
@@ -26,6 +26,9 @@ type List struct {
 	// The index of the currently selected item.
 	currentItem int
 
+	// The offset to ensure our currently selected item remains in view.
+	viewOffset int
+
 	// Whether or not to show the secondary item texts.
 	showSecondaryText bool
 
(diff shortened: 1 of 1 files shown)


rivo:
Totally agree with the requirement described in #217. However, here again, I'm providing my own implementation (see 0a17957), mostly because I find the use of two `switch` statements and the -1/+1/0 return values for the helper function a bit cumbersome. And it would have been quite a bit more code than what was needed.

So I'll close this one without merging. Hope you don't mind.

ardnew:
That -1/+1/0 is bad habit from my C programming, where that idiom is used pretty extensively, such as in `strcmp()` from the C standard library: 

`The strcmp() and strncmp() functions return an integer greater than, equal to, or less than 0, according as the string s1 is greater than, equal to, or less than the string s2.`

rivo:
I know `strcmp()`, C was one of the first languages I learned (25 years ago, haha!). That idiom makes sense where e.g. you need to sort an array, as there are three cases to consider (less, greater, equal). There are three cases here, too, but you're only using `contains()` once and only act on two of its cases. So I thought this could be simplified.

Thanks anyway for your submission. It's always interesting to see a different approach.

Issue #219, opened 2019-01-03
Add option to highlight entire row of selected List item


ardnew:
In some cases, it looks much better to have the entire row highlighted when navigating the list selections instead of only highlighting the item text itself, which will of course be variable length in many lists.

rivo:
Closed by #220

PR #220, opened 2019-01-03
Add option to highlight entire selected row of a List item

ardnew:
This implements the option requested in #219

### DIFF
--- a/list.go
+++ b/list.go
@@ -47,8 +47,8 @@ type List struct {
 	// If true, the selection is only shown when the list has focus.
 	selectedFocusOnly bool
 
-	// If true, the selected background color spans the entire view.
-	spanSelectionHighlight bool
+	// If true, the entire row is highlighted when selected.
+	highlightFullLine bool
 
 	// An optional function which is called when the user has navigated to a list
 	// item.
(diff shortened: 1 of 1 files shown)


rivo:
Generally ok with this. But could you please rename the member variable (and the function accordingly) to `highlightTextOnly`? I find `spanSelectedBackgroundColor` confusing.

GitHub says there are merge conflicts. You might want to pull in the latest changes before resubmitting.

rivo:
You know, I've been thinking about this some more and I think this should not be the default. It will look quite odd on very wide screens with little text on a line.

So I think this should be called `highlightFullLine` instead and one can then use `SetHighlightFullLine(true)` to activate it.

rivo:
Do you still want to make the proposed change? (It seems that you'll need to resolve a merge conflict, too, as the file has changed in the meantime.)

ardnew:
finally resolved conflict and updated the field name as requested, please pull when ready

ardnew:
> You know, I've been thinking about this some more and I think this should not be the default. It will look quite odd on very wide screens with little text on a line.

Yeah, I suppose in that case it might look odd. Is that really the most common case though?

To me it seems like the current implementation diverges from widespread convention. I can't think of any list in any application where only the text is highlighted (all context menus, dropdown menus, list boxes, combo boxes in macOS, Windows, Gnome, and KDE; all list controls in Chrome and Firefox). 

I realize these are graphical applications, but why would the convention be any different?

rivo:
I don't know. I just tried it on `demos/list/main.go` and it looks odd to me. And I don't even have a huge screen (working on a 13 inch laptop).

But if there are actual complaints from more users, I'm willing to rethink this.

PR #230, opened 2019-01-18
Added a reference to TableCell. Fixes feature request #229.

### LINKED ISSUE #229: Feature request: add reference to TableCell
`TableCell`s desperately needs a `reference` like in a `TreeView`https://github.com/rivo/tview/blob/master/treeview.go#L21
Otherwise it's so cumbersome to get back to the actual data when e.g. a cell or row was selected.

dguendisch:
Added a reference to the `TableCell` in the same line as it's already present for a `TreeNode`.
Fixes #229

### DIFF
--- a/table.go
+++ b/table.go
@@ -11,6 +11,9 @@ import (
 // directly but all colors (background and text) will be set to their default
 // which is black.
 type TableCell struct {
+	// The reference object.
+	Reference interface{}
+
 	// The text to be displayed in the table cell.
 	Text string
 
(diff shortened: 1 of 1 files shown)


rivo:
Any specific reason the `reference` field is private here? I think it can be a public field like the others.

dguendisch:
Actually no reason unless you had one in [`TreeNode#L21`](https://github.com/rivo/tview/blob/master/treeview.go#L21)
I just did it like there :)

rivo:
Can you make it public, please?

I realize that this is different from `TreeNode` where everything is private. In retrospect, I'd like all of these fields to be private but I started them as public fields and I want to remain backwards compatible. (`TreeView` was a late addition to the package.)

In this case, I'd like it to be consistent across one class so I think `TableCell.Reference` should be public. (I know `x`, `y`, and `width` are private but those are only internal members.)

dguendisch:
I see. Sure, made it public and rebased.

rivo:
Merged it, thanks!

PR #235, opened 2019-02-13
Prevent recompilation of regexes

Bios-Marcel:
I have noticed that there are some regular expressions in textview.go that get recompiled every time that they are needed.


rivo:
Thanks. I'll merge this after you've moved the "TabSize" comment into the right place. (Should be right before its declaration.)

rivo:
Merged it, thanks.

PR #247, opened 2019-03-07
Added Color getter in TreeNode

dgrr:


rivo:
Please name it `GetColor()` (similar to `GetText()` or `GetReference()`).

rivo:
Thanks!

PR #254, opened 2019-03-24
Support Go Modules

SamWhited:
I know you've had reservations about modules in the past, but last time you mentioned that you would merge modules support after it was broadly supported in Go.

Now that Go 1.12 is out, all supported versions of Go have full module support and use modules by default when a package isn't in the lelgacy GOPATH.
Please consider merging this updated `go.mod` file to make it easier on those who do use modules to consume this library.

Thanks for your consideration.


rivo:
I've added these myself. Thank you.

PR #256, opened 2019-03-31
Call the selected function on manual select

urandom:
On SetCurrentOption, call the main selected function if set, as well as
the option's selected function, if set.


rivo:
Thanks for the PR. I started to write some feedback and request some changes but it turns out the handling I'm envisioning for this required quite a bit more work, especially the function comments should be adapted accordingly. So I've gone ahead and made the change myself. Please see the referenced commit.

Issue #270, opened 2019-04-18
SetBackgroundColor(-1) issue


eu81273:
To match the background color with the terminal system background color, must set the SetBackgroundColor value to -1(tcell.ColorDefault).

Setting the BackgroundColor to -1 is works fine for most components, but for the List component produces unexpected results like below...

- When the list is updated, the updated list and the previous list are overlapped.
- When move the cursor down, the previous item remains in the selected background color.

diamondburned:
Try

```go
	app := tview.NewApplication()
	defer app.Stop()

	app.SetBeforeDrawFunc(func(s tcell.Screen) bool {
		s.Clear()
		return false
	})
```

eu81273:
@diamondburned 
Thank you. SetBeforeDrawFunc works perfect.

PR #271, opened 2019-04-19
Fix list ColorDefault issue #270

eu81273:
This PR is for #270 

**Bug Fix**

As described in #270, list have two problem with tcell.ColorDefault(-1) background.

- When the list is updated, the updated list and the previous list are overlapped.
- When move the cursor down, the previous item remains in the selected background color.

To fix this issue, I added background fields for drawing unselected item's background. (Is there any way to get Box's background value?)
And I modified main text printing logic to print with blank spaces for erase previous main text.

### DIFF
--- a/box.go
+++ b/box.go
@@ -242,11 +242,9 @@ func (b *Box) Draw(screen tcell.Screen) {
 
 	// Fill background.
 	background := def.Background(b.backgroundColor)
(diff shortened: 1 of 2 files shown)


eu81273:
Closed this PR thanks to [this solution](https://github.com/rivo/tview/issues/270#issuecomment-485083503)

PR #275, opened 2019-04-22
Added traceEnd=false when scrolling

thorsager:
I have *not* done a lot of research, so this might NOT be the right way of fixing the `.ScrollTo(row,column)` not-working-issue.

I found that the `TextView` sets `trackEnd=false` when scrolling up on `PgUP` but if you want to scroll based on _external_ event it is not working.

Anyway.. This patch fixed my issue, :)


rivo:
Sorry, I think I saw this just after I made this change myself.
Thanks for the PR anyway.

PR #280, opened 2019-05-10
Enable space to select tree node besides enter key.

linuxerwang:
Space is more frequently used for selection.


rivo:
I added the space key to `TreeView`, even though I would like to see data for your claim. (I for one never use space for selections.)

Thanks nonetheless for letting me know about this. If you want your PR's accepted in the future, please follow the [contribution guidelines](https://github.com/rivo/tview/blob/master/CONTRIBUTING.md).

PR #287, opened 2019-06-03
Added GetFormItems()

mikeschinkel:
As I could not find a current way to determine the number of form items from code written outside of the `tview.Form` class, the `GetFormItem(n)` method can't be used reliably if the number of fields is unknown.  For my use case I wanted to iterate all of them so I am submitting this pull request that would allow a developer to do so.

Hopefully this is simple and obvious enough of a change that it will be a no-brainer to commit.

### DIFF
--- a/form.go
+++ b/form.go
@@ -302,6 +302,12 @@ func (f *Form) GetFormItem(index int) FormItem {
 	return f.items[index]
 }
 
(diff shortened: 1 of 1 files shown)


mikeschinkel:
I am going to close this PR and create a new one because I accidentally created this one on the `master` branch my fork thus not allowing to me keep all PRs in my fork's `master` branch.

PR #293, opened 2019-06-04
Added a GetFocusedElement() method to Form

mikeschinkel:
I need to be able to display help based on which form field is current so I need to be able to access the current focused element for a help map.

Hopefully this is an acceptable PR?

### DIFF
--- a/form.go
+++ b/form.go
@@ -280,6 +280,14 @@ func (f *Form) ClearButtons() *Form {
 	return f
 }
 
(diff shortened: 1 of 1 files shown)


rivo:
I opted for  a different name, implementation, and comment. Thanks, though.

mikeschinkel:
@rivo — Good deal. The capability is the most important aspect for users of your package. This is a win. :-)

PR #296, opened 2019-06-07
Fixed TextView's reset [-] setting the wrong color

diamondburned:


senorprogrammer:
Looks like this PR is not necessary. The issue has been fixed in https://github.com/rivo/tview/blob/master/util.go#L125

PR #300, opened 2019-06-10
Added collection of FormItemArgs for each of the form FormItem types

mikeschinkel:
After working with Forms for a while and requesting several PRs to make working with them easier I finally had enough with the complexity and decided to implement an alternate method to add form fields than `<form>.AddInputField()`, `<form>.AddPasswordField()`, `<form>.AddDropdown()`,  and `<form>.AddCheckbox()`.

The *biggest problem I was trying to fix were that once a field was added it was tedious to then access the form to set other properties not allow to be set in the `FormItem`s `New...()` method.  So I tried numerous iterations with the **absolute requirement of backward compatibility** and came up with the following, by example:

```
type Object struct {
    Name     string
    Birthday string
    Password string
    Confirm  string
    FavColor string
    Website  string
    Customer bool
}
type Strings [] string

func (ss Strings) Index(str string) int {
    pos := -1
    for i, v := range ss {
        if v != str {
            continue
        }
        pos = i
        break
    }
    return pos
}

var colors = Strings{"Red", "Green", "Blue"}

func AddFields(form *tview.Form, obj Object) *tview.Form {
    return form.Clear(true).
        AddFormItem(NewInputField(), &InputFieldArgs{
            Label:       "Name:",
            Text:        obj.Name,
            FieldWidth:  40,
            Placeholder: "Enter your full name",
        }).
        AddFormItem(NewInputField(), &InputFieldArgs{
            Label:      "Birthday:",
            Text:       obj.Birthday,
            FieldWidth: 12,
            DoneFunc: func(key tcell.Key) {
                // do something with birthday
            },
        }).
        AddFormItem(NewDropDown(), &DropDownArgs{
            Label:         "Favorite Color?",
            Options:       colors,
            InitialOption: colors.Index(obj.FavColor),
            SelectedFunc: func(option string, optionIndex int) {
                // do something with selected option
            },
        }).
        AddFormItem(NewInputField(), &InputFieldArgs{
            Label:                "Website:",
            Text:                 obj.Website,
            FieldWidth:           60,
            FieldBackgroundColor: tcell.ColorAliceBlue,
            FieldTextColor:       tcell.ColorGreen,
        }).
        AddFormItem(NewCheckbox(), &CheckboxArgs{
            Label:   "Customer?",
            Checked: obj.Customer,

        }).
        AddFormItem(NewPasswordField(), &PasswordFieldArgs{
            Label:      "Password:",
            Text:       obj.Password,
            FieldWidth: 30,
        }).
        AddFormItem(NewPasswordField(), &PasswordFieldArgs{
            Label:      "Confirm:",
            Text:       obj.Confirm,
            FieldWidth: 30,
        })
}
```
I would have preferred to use the names `AddInputField()` etc. but could not do so w/o breaking old code so I extended `AddFormItem()` instead.

There is also a ton of effectively duplicated code, but given the structure of the existing code I could not find a way to consolidate code without breaking backward compatibility.  If breaking backward compatibility is an option, I could change the code to be more maintainable, but I needed the functionality and I assumed you would be much less likely to access broken backward compatibility so I went ahead and implemented as is.

Hopefully you will concur this is a good addition to tview and accept this PR?  Thanks in advance for considering.

### DIFF
--- a/checkbox.go
+++ b/checkbox.go
@@ -201,3 +201,96 @@ func (c *Checkbox) InputHandler() func(event *tcell.EventKey, setFocus func(p Pr
 		}
 	})
 }
+
+// CheckboxArgs provides a concise and readable way to initialize
+// all the properties of the Checkbox struct by passing to the
+// <form>.AddFormItem(item,args).
+type CheckboxArgs struct {
+	baseFormItemArgs
+
+	// The text to be displayed before the input area.
+	Label string
+
+	// Whether or not this box is checked.
+	Checked bool
+
+	// An optional function which is called when the user changes
+	// the checked state of this checkbox.
+	ChangedFunc func(checked bool)
+
+	// The screen width of the input area. A value of 0 means extend
+	// as much as possible.
+	FieldWidth int
+
+	// The screen width of the label area. A value of 0 means use
+	// the width of the label text.
+	LabelWidth int
+
+	// The label color.
+	LabelColor tcell.Color
+
+	// The background color of the input area.
+	FieldBackgroundColor tcell.Color
+
+	// The text color of the input area.
+	FieldTextColor tcell.Color
+
+	// The background color of the input area.
+	BackgroundColor tcell.Color
+
+	// An optional function which is called when the user indicated
+	// that they are done entering text. The key which was pressed
+	// is provided (tab, shift-tab, enter, or escape).
+	DoneFunc func(key tcell.Key)
+
+	// A callback function set by the Form class and called when
+	// the user leaves this form item.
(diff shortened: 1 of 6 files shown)


mikeschinkel:
@rivo:

I wanted to follow up with some links showing this feature in action:

1. https://github.com/newclarity/blueprintz/blob/master/browseui/project_view.go#L76
2. https://github.com/newclarity/blueprintz/blob/master/browseui/core_view.go#L30
3. https://github.com/newclarity/blueprintz/blob/master/browseui/layout_view.go#L41

PR #308, opened 2019-06-16
Don't throw away nodes after drawing

ollien:
Currently, a tree's nodes are thrown away after drawing a `TreeView`, and then reprocessed on every Draw. This causes `GetRowCount` to always return 0 from the `TreeView`. This PR fixes that problem by simply preserving the nodes between draws.


rivo:
Thanks.

PR #319, opened 2019-06-30
Sleep presentation TextView1 example when inactive

ajcarvajal:
A goroutine in the demo presentation increments an integer infinitely in the background. Now checks hasFocus() and sleeps longer if inactive

### DIFF
--- a/demos/presentation/textview.go
+++ b/demos/presentation/textview.go
@@ -45,9 +45,13 @@ func TextView1(nextSlide func()) (title string, content tview.Primitive) {
 	go func() {
 		var n int
 		for {
-			n++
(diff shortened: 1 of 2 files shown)


rivo:
The demo presentation was never meant to be run for a longer time so I'm not sure this is a real problem. But it's ok. I'll merge your commit if you exclude the "formatting fixes" in treeview.go. (I'd like to keep it brief to provide a good overview and gofmt is still fine with it.)

ajcarvajal:
Cool, didn't mean to include the formatting changes anyway. Thanks for taking a look. I'm trying to get involved as a new dev, so it's appreciated!

PR #328, opened 2019-07-26
Use Lock to Clear() TextView

khepin:
Currently it's possible to call `Clear()` on a TextView while it is currently `Draw()`ing.
This would result in the view having an empty buffer when it is iterating through the
lines on line 839:

`text := t.buffer[index.Line][index.Pos:index.NextPos]`

This will then cause a panic and crash the whole program.

Since `Draw()` already locks, making sure that `Clear()` also does will prevent
the buffer from disappearing in the middle of a draw.

### DIFF
--- a/textview.go
+++ b/textview.go
@@ -365,6 +365,8 @@ func (t *TextView) GetScrollOffset() (row, column int) {
 
 // Clear removes all text from the buffer.
 func (t *TextView) Clear() *TextView {
+	t.Lock()
+	defer t.Unlock()
 	t.buffer = nil
 	t.recentBytes = nil
 	t.index = nil


rivo:
This is by design — or rather by necessity. There are lots of other functions which are also not thread-safe. See [the documentation](https://godoc.org/github.com/rivo/tview#hdr-Concurrency):

> Many functions in this package are not thread-safe. [...] If you access your primitives from other goroutines, however, you will need to synchronize execution.

For a more detailed treatment of the topic, please refer to the Wiki:

https://github.com/rivo/tview/wiki/Concurrency

khepin:
Oh, I had completely missed that part of the doc!!! Very helpful, thanks.

Issue #225, opened 2019-01-16
Feature request: add title or text Spinner (example included)


dguendisch:
Would be cool to have a title or text Spinner for e.g. a `Box` indicating that a (potentially long) operation is currently being executed. Maybe a more basic primitive would be even more suitable (e.g. to have Spinner in a status bar, etc).

![2019-01-16_22-54-58](https://user-images.githubusercontent.com/30311254/51281243-1eb51d00-19e2-11e9-92bf-07ccc59a184b.gif)

I currently get by myself with this simple function:
```
func main() {
    app := tview.NewApplication()
    box := tview.NewBox().SetBorder(true).SetTitle("Hello")

    spinTitle(app, box, "Hello", func() {
        //something long running
        time.Sleep(3 * time.Second)
    })

    if err := app.SetRoot(box, true).Run(); err != nil {
        log.Panic("Couldn't construct tview application")
    }
}

// spinTitle will execute action and prepend the box's title with a spinning character until action terminates
func spinTitle(app *tview.Application, box *tview.Box, title string, action func()) {
    done := make(chan bool)

    //action
    go func() {
        action()
	done <- true
        close(done)
    }()

// spinner
    go func() {
        spinners := []string{"⠋", "⠙", "⠹", "⠸", "⠼", "⠴", "⠦", "⠧", "⠇", "⠏"}
	var i int
	for {
	    select {
	    case _ = <-done:
	        app.QueueUpdateDraw(func() {
		    box.SetTitle(title)
		})
		return
	    case <-time.After(100 * time.Millisecond):
		spin := i % len(spinners)
		app.QueueUpdateDraw(func() {
	            box.SetTitle(spinners[spin] + title)
		})
		i++
            }
        }
    }()
}
```

Sacules:
I'd be very interested in something like this :)

rivo:
There was a similar request previously for a progress bar, see #83, but it sounds like this is not what you are looking for.

I agree this would be useful. I'm just not yet sure what this should look like. I don't think it should become part of the title as in your example. Maybe the bottom of a box frame could be used to display the status?

Any suggestions?

dguendisch:
Yeah, it's just a busy indicator (e.g. doing a network request and waiting for the result), so a progress bar would be hard to "progress" and overkill in terms of screen size :)
[...]

PR #331, opened 2019-08-01
Add footer to Box

shanempope:
I needed this for my own project. Feel free to merge or close.

### DIFF
--- a/box.go
+++ b/box.go
@@ -44,6 +44,15 @@ type Box struct {
 	// The alignment of the title.
 	titleAlign int
 
+	// The footer. Only visible if there is a border, too.
+	footer string
+
+	// The color of the footer.
+	footerColor tcell.Color
+
+	// The alignment of the footer.
+	footerAlign int
+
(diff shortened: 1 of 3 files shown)


shanempope:
image

rivo:
Thanks.

Instead of adding a footer, I'd rather add a general "decoration" slice to `Box` with parameters such as text, vertical alignment (left, middle, right), and horizontal alignment (also left, middle, right).

Mostly because, next thing you know, somebody will ask for ways to add something in the left border or the top right corner. Or, say, two titles (one aligned left, one aligned right).

This would also be a way to deal with #225. But since I'm even having trouble catching up with all the PR's sent to me, I'm not sure when I will get to that. But I'll keep this PR open and will close it once the decoration stuff is available.

In the meantime, one can always use [`Box.SetDrawFunc()`](https://godoc.org/github.com/rivo/tview#Box.SetDrawFunc) to add more custom elements to a primitive.

PR #336, opened 2019-08-24
Don't skip regions at end of line

SamWhited:
Currently if a region is the last thing on a line, it is skipped. eg. if the "highlight" region is highlighted, the entire text below will be highlighted even though it should have ended at the next region:

```
["highlight"]Test[""]

This is highlighted too.
```

This fixes an off-by-one issue that appears to have been causing the problem.

### DIFF
--- a/textview.go
+++ b/textview.go
@@ -662,7 +662,7 @@ func (t *TextView) reindexBuffer(width int) {
 				}
 
 				// Is the next tag in range?
-				if tagIndex < 0 || minPos >= tagEnd+remainingLength {
+				if tagIndex < 0 || minPos >= tagEnd+remainingLength+1 {
 					break // No. We're done with this line.
 				}
 


SamWhited:
Gentle ping, this bug is still affecting me. Thanks!

Bios-Marcel:
You could make a fork and redirect rivo/tview to SamWhited/tview using go modules. Just sayin ~

SamWhited:
> You could make a fork and redirect rivo/tview to SamWhited/tview using go modules. Just sayin ~

That's what I'm doing right now, but replace directives are not transitive so it's a temporary solution at best. I just like to ping every week or two on bug fixes and what not to make sure they don't get lost in the backlog :)

rivo:
Sorry for the delay on this.

The fix is not your PR because it should be `>` instead of `>=`, not a `+1` at the end.

PR #337, opened 2019-08-25
Feat: add Application.Set(Get)UnknownEventHandler

dzpao:
Add Application.Set(Get)UnknownEventHandler for handling events that tview
does not yet support, such as a mouse event or an user defined event.

### DIFF
--- a/application.go
+++ b/application.go
@@ -43,6 +43,10 @@ type Application struct {
 	// be forwarded).
 	inputCapture func(event *tcell.EventKey) *tcell.EventKey
 
+	// An optional handler for handling events that tview does not yet support,
+	// such as a mouse event handler or a custom event handler.
+	unknownEventHandler func(event tcell.Event)
+
 	// An optional callback function which is invoked just before the root
 	// primitive is drawn.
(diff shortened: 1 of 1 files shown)


application.go
```diff
@@ -246,6 +263,15 @@ EventLoop:
...
 				screen.Clear()
 				a.draw()
+			default:
+				a.RLock()
+				unknownEventHandler := a.unknownEventHandler
+				a.RUnlock()
+
+				if unknownEventHandler != nil {
+					unknownEventHandler(event)
+					continue // Don't forward event.
```
tslocum: Is this necessary?  The only effect of this seems to be skipping the chance of calling ```updater```.

millerlogic:
Should this be closed for the same reason #338 was closed?
Specifically https://github.com/rivo/tview/pull/338#issuecomment-549189901 and the workaround code provided.

rivo:
I think so, yes. Mouse handling will come and I don't like the idea of some blanket event handler. I'd like to see an actual problem that results from not having this first.

PR #338, opened 2019-08-25
Feat: add limited mouse support

dzpao:
Add Application.EnableMouse() to enable tcell mouse support.
A limited mouse support is available with #337.

### DIFF
--- a/application.go
+++ b/application.go
@@ -43,6 +43,10 @@ type Application struct {
 	// be forwarded).
 	inputCapture func(event *tcell.EventKey) *tcell.EventKey
 
+	// An optional handler for handling events that tview does not yet support,
+	// such as a mouse event handler or a custom event handler.
+	unknownEventHandler func(event tcell.Event)
+
 	// An optional callback function which is invoked just before the root
 	// primitive is drawn.
 	beforeDraw func(screen tcell.Screen) bool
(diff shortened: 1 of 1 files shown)


millerlogic:
Hi, what would make me hesitate to use this is the fact that "unknown" events aren't well defined and could change in the future, thus breaking compatibility. Instead, it could be raw event for all events, or mouse event just for mouse events. I think I like the idea of a raw event handler, it preprocesses all events, and could allow you to intercept and change behavior; and in this case, to receive mouse events.

millerlogic:
It's actually fairly easy to do this without making any changes, wrap the Screen and grab the events:

```go
type MouseScreen struct {
	tcell.Screen
	MouseEventHandler func(event *tcell.EventMouse)
}

func (screen *MouseScreen) PollEvent() tcell.Event {
	event := screen.Screen.PollEvent()
	if mouseEvent, ok := event.(*tcell.EventMouse); ok {
		screen.MouseEventHandler(mouseEvent)
	}
	return event
}
```
Use:

```go
	app := tview.NewApplication()
	tscreen, err := tcell.NewScreen()
	if err != nil {
		return err
	}
	screen := &MouseScreen{Screen: tscreen,
		MouseEventHandler: func(event *tcell.EventMouse) {
			// ... handle mouse event here ...
		}}
	err = screen.Init()
	if err != nil {
		return err
	}
	screen.EnableMouse()
	app.SetScreen(screen)
```

dzpao:
Thanks @millerlogic, I learned a lot from here.

PR #342, opened 2019-08-31
Fix: non-scrollable textview has strange behavior

### LINKED ISSUE #341: The behavior of TextView.SetScrollable() is very strange
Talk is cheap, show you the code:

```Golang
package main

import (
	"fmt"
	"time"

	"github.com/rivo/tview"
)

func main() {
	app := tview.NewApplication()
	textView := tview.NewTextView()
	textView.SetScrollable(false)  // <------ Please note here
	textView.SetBorder(true)
	textView.SetChangedFunc(func() {
		text := textView.GetText(false)
		textView.SetTitle(fmt.Sprintf("%d bytes", len(text)))
		app.Draw()
	})

	go func() {
		n := 1
		for {
			fmt.Fprintf(textView, "line %d\n", n)
			n++
			time.Sleep(200 * time.Millisecond)
		}
	}()

	if err := app.SetRoot(textView, true).Run(); err != nil {
		panic(err)
	}
}
```

When I ran this code, I observed that the screen would be empty after it was full.
But according to the TextView documentation, it should be that only text that is above the top of the screen will be discarded.

>    If the text is not scrollable, any text above the top visible line is discarded.

So what am I doing wrong?

dzpao:
It seems that there are two lines of code that are upside down.

This should fix #341

### DIFF
--- a/textview.go
+++ b/textview.go
@@ -940,9 +940,9 @@ func (t *TextView) Draw(screen tcell.Screen) {
 	// scrolled out of view.
 	if !t.scrollable && t.lineOffset > 0 {
 		if t.lineOffset <= len(t.index) {
-			t.buffer = nil
-		} else {
 			t.buffer = t.buffer[t.index[t.lineOffset].Line:]
+		} else {
+			t.buffer = nil
 		}
 		t.index = nil
 		t.lineOffset = 0


dzpao:
In order to explain the problem more clearly, I wrote a short example.

If you run the example with the master version, you can see that every time the screen is full, it will be emptied.

With patch applied, the screen will continue to scroll up without being emptied. I think this is more in line with people's expectations.

```Go
package main

import (
	"fmt"
	"time"

	"github.com/rivo/tview"
)

func main() {
	app := tview.NewApplication()
	textView := tview.NewTextView().
		SetScrollable(false).
		SetChangedFunc(func() {
			app.Draw()
		})

	go func() {
		i := 0
		for {
			fmt.Fprintf(textView, "%d\n", i)
			i++
			time.Sleep(200 * time.Millisecond)
		}
	}()

	textView.SetBorder(true)
	mainView := tview.NewFlex().
		SetDirection(tview.FlexRow).
		AddItem(textView, 10, 1, true)
	if err := app.SetRoot(mainView, true).Run(); err != nil {
		panic(err)
	}
}
```

PR #343, opened 2019-09-09
Feat: allow to override key bindings for tree actions

jumale:
## Previous behaviour
The `treeview` provides some statically hardcoded key bindings for interacting with the tree. The consumers of the package are not able to re-define those bindings.

## New behaviour
The `treeview` reads the key bindings from a configurable map. 
There is a default map of keys to actions, initialised for each new `TreeView`.
However it's possible to redefine key bindings for each action, using method `SetKeyBinding`

## Benefits
I want to be able to reassign the default key bindings and implement some custom tree actions for some keys which currently taken by treeview.

### DIFF
--- a/demos/treeview/main.go
+++ b/demos/treeview/main.go
@@ -56,6 +56,11 @@ func main() {
 		}
 	})
 
+	// Example of overriding a key binding to select a tree node by pressing G
+	tree.SetKeyBinding(tview.TreeSelectNode, 'G')
+	// Example of overriding a key binding to move up by pressing "up" or "left"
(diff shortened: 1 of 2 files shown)


rivo:
First of all, apologies for the late reply. I'm going to close this one.

> The consumers of the package are not able to re-define those bindings.

They are. Please see [`SetInputCapture()`](https://godoc.org/github.com/rivo/tview#Box.SetInputCapture) for details. I currently don't see the need to have a key map in there, especially not as a `TreeView`-only solution. I purposely made `tview` a "opinionated" to get users started quickly. To change that behaviour, there are intercept functions such as `SetInputCapture()`.

PR #344, opened 2019-09-12
Refactor: make the correct processing of SGR code

### LINKED ISSUE #320: ANSIWriter doesn't support SGR 2;37;0m
`ANSIWriter` doesn't support SGR `ESC [2;37;0m`. This is a very popular sequence (sorry for can't give the standard which it follows).

dzpao:
* SGR code 21/22/23/24/25/27 should not turn off all attributes.
* turn on/off one of bold/dim/underline/blink should not turn off other attributes.

This should fix #320

### DIFF
--- a/ansi.go
+++ b/ansi.go
@@ -90,17 +90,21 @@ func (a *ansi) Write(text []byte) (int, error) {
 					fmt.Fprint(a.buffer, strings.Repeat("\n", count))
 				case 'm': // Select Graphic Rendition.
 					var (
-						background, foreground, attributes string
-						clearAttributes                    bool
+						background, foreground                       string
+						reset                                        bool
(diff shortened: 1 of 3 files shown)


antonmedv:
@rivo, Hi. Can you please take a look at this?

tslocum:
Thanks, this looks good but I have a few thoughts:

If tcell doesn't support italics yet, leaving support for it out entirely seems best.

Would an enum or int (-1, 0, or 1) be easier to read/understand than a pointer to bool?

A new type and method might allow clearer attribute state management and encoding:

```go
type Attributes struct {
    reset bool
    bold, dim, italic, underline, blink, reverse *bool
}

func (a *Attributes) String() string {
    // Return attribute string
}
```

rivo:
See #320 for comments.

PR #347, opened 2019-09-14
Handle ansi code 39 and 49

Seanstoppable:
39 is apparently the "default foreground color"
49 is apparently the "default background color"

Without change:

Screen Shot 2019-09-13 at 8 19 07 PM

With change:

Screen Shot 2019-09-13 at 8 14 47 PM


rivo:
Thanks for letting me know. I'm not using this PR, though, because of a few issues. It ignores the definition of the colour tags in `tview` (specifically the use of `-` to reset colours). The color "default" does not exist. Lastly, it should have turned up a bug in the implementation of `overlayStyle()`.

The latest commit incorporates all of these.

Seanstoppable:
Got it. Appreciate the update!

Issue #345, opened 2019-09-13
Table: be able to fix the width of each column when scrolling


cespedes:
In some cases, it would be nice to have the width of all the columns in a Table fixed, so that when using a large table with cells of different size and scroll through it, every column is always in the same place.

I needed this functionality for one of my projects, so I already forked tview and patched it in https://github.com/cespedes/tview adding a SetFixedColumnsWidth(bool) to *Table

It would be nice if you could merge it (or implement it in any other way), so that I could use the main repository instead of a fork.

Thank you,

cespedes:
After reviewing the issues, I see this one is similar to #119 but I still think it would be good if the package could provide one way to fix the columns width...

rivo:
Thanks. I added this. It has a different name, though. See the referenced commit.

PR #348, opened 2019-09-23
Table.SetFixedColumnsWidth():  keep columns width fixed when scrolling

cespedes:
Added a new method to Table, SetFixedWidth(), to calculate the maximum width of every column and maintaining it fixed, even if the displayed rows have smaller widths.

This will prevent the tables from changing columns width when scrolling.

This fixes issue #345

### DIFF
--- a/demos/table/main.go
+++ b/demos/table/main.go
@@ -11,6 +11,7 @@ import (
 func main() {
 	app := tview.NewApplication()
 	table := tview.NewTable().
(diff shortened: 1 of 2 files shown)


rivo:
This should already be resolved with [`Table.SetEvaluateAllRows()`](https://godoc.org/github.com/rivo/tview#Table.SetEvaluateAllRows).

PR #353, opened 2019-10-14
Add window size change handler

tslocum:
My thanks to the tcell & tview contributors!  The migration has been a pleasure.

### DIFF
--- a/application.go
+++ b/application.go
@@ -43,6 +43,10 @@ type Application struct {
 	// be forwarded).
 	inputCapture func(event *tcell.EventKey) *tcell.EventKey
 
+	// An optional callback function which is invoked when the application's
+	// window is initialized, and when the application's window size changes.
+	afterResize func(screen tcell.Screen)
+
 	// An optional callback function which is invoked just before the root
 	// primitive is drawn.
 	beforeDraw func(screen tcell.Screen) bool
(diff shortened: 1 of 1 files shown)


rivo:
I'm generally ok with this but I wonder why we need to supply the callback with the `Screen` object? This is likely not the right time for a client to draw something onto the screen. (The documentation also does not say what state the screen is in. Has it already been redrawn? Or will it be redrawn after the callback is invoked?) For redraw events, there is [`SetBeforeDrawFunc`](https://godoc.org/github.com/rivo/tview#Application.SetBeforeDrawFunc) and [`SetAfterDrawFunc()`](https://godoc.org/github.com/rivo/tview#Application.SetAfterDrawFunc).

If information such as the new size of the screen is needed, we should provide that directly instead.

What do you think?

tslocum:
Thanks.  I agree that providing width/height directly is best.  I've updated this PR with that change as well as clarifying that the screen is cleared and the application is redrawn after invoking this callback.

application.go
```diff
@@ -241,6 +247,13 @@ EventLoop:
 				a.RLock()
 				screen := a.screen
 				a.RUnlock()
+
+				// Call afterResize handler if there is one.
+				if a.afterResize != nil {
+					width, height := screen.Size()
```
rivo: This will panic if `screen` is `nil` (which can happen, as can be seen further below). Can you move this block to after the `screen == nil` check?

tslocum: Thanks. PR updated.

PR #358, opened 2019-10-30
Added CustomKeys support in TextView

aravindgopall:
In Current implementation, InputHandler function it is hardcoded when to call function `done` which is defined by user. In new implementation added a key customKeys to
TextView which can be set by passing `[]tcell.Key` and handle them inside
user defined function `done`.

Later we can remove the other three functions and make it completely user defined keys.

### DIFF
--- a/demos/textview/main.go
+++ b/demos/textview/main.go
@@ -21,13 +21,15 @@ Capitalize on low hanging fruit to identify a ballpark value added activity to b
 
 func main() {
 	app := tview.NewApplication()
+	customKeys := []tcell.Key{tcell.KeyF1}
 	textView := tview.NewTextView().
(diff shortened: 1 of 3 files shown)


rivo:
If you need to handle or translate any custom keys, please use [`SetInputCapture()`](https://godoc.org/github.com/rivo/tview#Box.SetInputCapture). You should be able to achieve what you describe with this function.

Because this topic comes up every now and then, I've created a Wiki page about custom keys: https://github.com/rivo/tview/wiki/CustomKeys

I'm also closing this PR because it goes against many of the points listed in the [contributing guide](https://github.com/rivo/tview/blob/master/CONTRIBUTING.md).

Issue #365, opened 2019-11-11
Iteration difficult on form values.


stephen-mw:
It's possible I'm missing something obvious for a retrieving a list of `items` in `Forms`. Currently the `f.items` array isn't externally accessible, which means when I need to for-loop over the form items via `f.GetFormItem(N)` and catch the out-of-index panic (or, keep a secondary index of things added to the form).

If I'm not missing something obvious, please accept my PR, which makes it easier to do loops on the items by exposing the `len` of available form items.

stephen-mw:
Looks like there's an existing fix in PR 298

PR #366, opened 2019-11-11
Add function for form item counts.

stephen-mw:
Fix for issue #365 

Add getter for len of available form items.


stephen-mw:
Looks like there's an existing fix in PR 298

PR #368, opened 2019-11-14
Add support for displaying text next to a checkbox

berrange:
When building forms the label field is typically quite short, just one
or two words. For checkboxes it is often desirable to have a longer
descriptive piece of text. This is not practical to use as a label and
in many applications would more commonly be placed to the right of the
checkbox.

This adds support for a "SetMessage()" method which provides support
for text adjacent to the checkbox

As an example, this form shows one usage pattern, where the checkbox
is used to require case sensitive matching of the author name query.
In this case the checkbox label is an empty string, and the message
text is used instead:

```
 ╔════════════ User filtering ════════════════════╗
 ║                                                ║
 ║    Age:   ________                             ║
 ║                                                ║
 ║ Author:   ______________                       ║
 ║                                                ║
 ║           X Case sensitive                     ║
 ║                                                ║
 ║   Apply     Cancel                             ║
 ╚════════════════════════════════════════════════╝
```

Another pattern is where there are a series of checkboxes, all
related to the same attribute. Thus they have a common form
label but different message text

```
 ╔════════════ Request filtering ═════════════════╗
 ║                                                ║
 ║    State:    X Opened                          ║
 ║                                                ║
 ║              _ Closed                          ║
 ║                                                ║
 ║              _ Merged                          ║
 ║                                                ║
 ║              X Locked                          ║
 ║                                                ║
 ║   Apply     Cancel                             ║
 ╚════════════════════════════════════════════════╝
```

Signed-off-by: Daniel P. Berrangé <berrange@redhat.com>

### DIFF
--- a/checkbox.go
+++ b/checkbox.go
@@ -17,6 +17,9 @@ type Checkbox struct {
 	// The text to be displayed before the input area.
 	label string
 
+	// The text to be displayed after the checkbox
+	message string
+
 	// The screen width of the label area. A value of 0 means use the width of
 	// the label text.
 	labelWidth int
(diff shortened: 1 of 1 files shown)


tslocum:
Tested and merged into [cview](https://git.sr.ht/~tslocum/cview) with [this additional commit](https://git.sr.ht/~tslocum/cview/commit/c90e430fdfdad1d5584765f46d34d88044113785) adding a message argument to Form.NewCheckbox

berrange:
Updated the patch to apply to current git master

PR #369, opened 2019-11-15
Guards against a panic in dropdown.SetCurrentOption(-1)

ewollesen:
When SetCurrentOption is called, that option's Selected hook is called,
if present. However, should -1 (no selection) be passed, then the code
would panic with an index out of range error.


rivo:
Thank you. But shouldn't it also be checked in the lines before, i.e. for line [435](https://github.com/rivo/tview/blob/023b522d351167c0a085c82c890dbaf1d835b521/dropdown.go#L435)?

PR #389, opened 2020-01-17
Add Application.RingBell

tslocum:
This rings the terminal bell on Unix and is untested on Mac/Windows.


rivo:
Does this really need a lock? We're not accessing any local variables.

If this can conflict with any other output to the terminal, it should rather be queued with `QueueUpdate()`. I'm not familiar enough with the bell character. Can it interrupt any other character sequences if output at the wrong time?

tslocum:
Thanks. I also don't know the possible extent of corruption. I've updated this to use QueueUpdate instead.

PR #392, opened 2020-01-21
Add Flex.Clear() method to remove all items in flexbox

blefevre:
Other components that contain a list of items have a `Clear()` method to remove them all.
E.g:
* [Grid.Clear()](https://godoc.org/github.com/rivo/tview#Grid.Clear)
* [List.Clear()](https://godoc.org/github.com/rivo/tview#List.Clear)
* [Table.Clear()](https://godoc.org/github.com/rivo/tview#Table.Clear)

This PR adds the same `Clear()` method to the Flex component.

### DIFF
--- a/flex.go
+++ b/flex.go
@@ -101,7 +101,7 @@ func (f *Flex) RemoveItem(p Primitive) *Flex {
 	return f
 }
 
-// Clear removes all items from the flexbox
+// Clear removes all items from the container.
 func (f *Flex) Clear() *Flex {
 	f.items = nil
 	return f
--- a/flex.go
+++ b/flex.go
@@ -101,6 +101,12 @@ func (f *Flex) RemoveItem(p Primitive) *Flex {
 	return f
 }
 
+// Clear removes all items from the flexbox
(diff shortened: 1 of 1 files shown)


tslocum:
This method is also missing from Pages.

flex.go
```diff
@@ -101,6 +101,12 @@ func (f *Flex) RemoveItem(p Primitive) *Flex {
 	return f
 }
 
+// Clear removes all items from the flexbox
```
rivo: Can you change this to the following, please?

> Clear removes all items from the container.

(Notice the period at the end.)

It may seem like a small and irrelevant change but I kept  the entire package consistent in this regard and would like to keep it this way. See also the [contributing guidelines](https://github.com/rivo/tview/blob/master/CONTRIBUTING.md):

> Start all sentences upper-case, as is common in English, and _end them with a period._

rivo:
Thanks. I added a comment with a small change request.

If you want, please go ahead and include the `Clear()` function for `Pages`.

PR #395, opened 2020-01-28
Always fill Box background

tslocum:
This resolves rendering issues with TextViews which have their background
color set to ColorDefault.

### DIFF
--- a/box.go
+++ b/box.go
@@ -255,11 +255,9 @@ func (b *Box) Draw(screen tcell.Screen) {
 
 	// Fill background.
 	background := def.Background(b.backgroundColor)
-	if b.backgroundColor != tcell.ColorDefault {
-		for y := b.y; y < b.y+b.height; y++ {
-			for x := b.x; x < b.x+b.width; x++ {
-				screen.SetContent(x, y, ' ', nil, background)
-			}
+	for y := b.y; y < b.y+b.height; y++ {
+		for x := b.x; x < b.x+b.width; x++ {
+			screen.SetContent(x, y, ' ', nil, background)
 		}
 	}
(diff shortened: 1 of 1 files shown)


tslocum:
Without this change, when setting the content of a TextView with its background set to ColorDefault to a shorter length than it previously was, anywhere not drawn upon with new text will continue displaying the previous text.

rivo:
Can you give an example where `ColorDefault` is a good idea for the background colour of a `TextView`?

tslocum:
Setting the foreground and background to ColorDefault allows the usage of users' default terminal colors.  Filling with a solid color instead may affect terminal transparency (Alacritty+Linux) and/or not combine aesthetically with users' terminal colors, as they are unknown.  It's best to have a choice of whether or not to specify a set of colors to use, which this PR allows.  Using the colors chosen by the user is simple:

```go
tview.Styles.TitleColor = tcell.ColorDefault
tview.Styles.BorderColor = tcell.ColorDefault
tview.Styles.PrimaryTextColor = tcell.ColorDefault
tview.Styles.PrimitiveBackgroundColor = tcell.ColorDefault
```

ajisaac:
I also have this issue, where I want the background to use ColorDefault to have a semi transparent background, but then it won't update those portions of the widget.

problame:
Also hit this issue, with the same motivation as @tslocum : the tui we're building ( https://github.com/zrepl/zrepl/pull/297 ) is monochromatic and thus we want to use the user's default terminal emulator colors. The effect for the end user is that the tui feels like it automatically respects the user's preference for light / dark terminal emulator theme.

PR #403, opened 2020-02-13
Fix scrolling table with PageDown and PageUp

### LINKED ISSUE #402: Table pageUp & pageDown ignores fixedRows
With a table longer than one page, if you page up or down with fixedRows set to 1 or greater, the text moves up or down the page size including the number of fixedRows, so if you are paging through values, you'll miss out on seeing one (or more) values. For instance if the current number of rows is 1 fixed + 6 non-fixed, the page up/down will move 7 rows when it should (perhaps) only move 6.

You may want to adjust the value of visibleRows in the Draw func when fixedRows is not zero, or perhaps adjust for fixedRows in the InputHandler in the pageUp/pageDown func.

tslocum:
Resolves #402.

### DIFF
--- a/table.go
+++ b/table.go
@@ -1101,27 +1101,37 @@ func (t *Table) InputHandler() func(event *tcell.EventKey, setFocus func(p Primi
 			}
 
 			pageDown = func() {
+				offsetAmount := t.visibleRows - t.fixedRows
+				if offsetAmount < 0 {
+					offsetAmount = 0
+				}
+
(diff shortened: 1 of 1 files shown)


rivo:
Thank you. This is generally ok but I would like you to handle the  case  where `t.visibleRows < t.fixedRows`. Scrolling should then not be reversed. There should be no scrolling at all.

Could you please add that?

tslocum:
Thanks.  PR updated.

PR #404, opened 2020-02-13
Clear screen when resizing

### LINKED ISSUE #401: [Windows] Resize event not sending updated width and height
Compiling any of the demos on windows and trying to run the exe both directly and through cmd, powershell, and Windows Terminal, the UI does not refresh after resizing at all!

I have to manually press a key for the UI to refresh.

Editing `application.go` and adding `screen.Sync()` before `newWidth, newHeight := screen.Size()` seems to fix it?

Not sure exactly why this issue is happening?

tslocum:
Resolves #401.

### DIFF
--- a/application.go
+++ b/application.go
@@ -132,7 +132,6 @@ func (a *Application) SetScreen(screen tcell.Screen) *Application {
 func (a *Application) Run() error {
 	var (
 		err           error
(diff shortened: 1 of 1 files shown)


rivo:
I made this change myself because for some reason, there were merge conflicts so it was easier to just do it myself.

PR #423, opened 2020-03-30
implement modal_input - like js 'prompt' primitive

holiman:
This PR implements another modal primitive, similar to javascript `prompt` method. I'm a bit new to tview, and I'm not quite sure if things like this belong in the upstream library or if they are expected to be written in the upper layer. However, it seemed natural to make it similar to the existing `modal` windows. Please let me know if I've missed some easier way to achieve the same goal.


rivo:
Please read https://github.com/rivo/tview/blob/master/CODE_OF_CONDUCT.md before submitting PRs. Specifically, there was no discussion whether this is really needed so it's unlikely I'm merging this.

Of course, one can imagine a large variety of modals but in my opinion, it makes no sense to create a class for each of them. [This Wiki page](https://github.com/rivo/tview/wiki/Modal) explains how to create a modal window from other primitives. If you need an input field, `Form` would be a good start as it also has buttons.

Unless there's a good argument otherwise, I'll close this PR soon.

holiman:
> Please read https://github.com/rivo/tview/blob/master/CONTRIBUTING.md before submitting PRs. Specifically, there was no discussion whether this is really needed so it's unlikely I'm merging this.

Sorry, in my mental model I assumed that a PR was a good base for discussion. I implemented it so I could get a better understanding of tview in general, and I don't at all mind if you close this. 

I'll look more closely into how to build this from the examples provided instead.

rivo:
Ok, I understand. In _my_ mental model, when I get sent PRs, it means "please review my code and merge if possible". Because many PRs appear to solve only niche requirements of some users and I don't have time to review all of that code, I prefer to discuss the need for those features/changes first, and an issue is the right place for that.

PR #431, opened 2020-04-17
Allow changing current node and changing focus in treeview handlers

millerlogic:
Without this, SetCurrentNode(nil) in a handler is either ignored or panics on nil.
Also I'd like to change focus on selected.

### DIFF
--- a/treeview.go
+++ b/treeview.go
@@ -747,13 +747,14 @@ func (t *TreeView) MouseHandler() func(action MouseAction, event *tcell.EventMou
 			if y >= 0 && y < len(t.nodes) {
 				node := t.nodes[y]
 				if node.selectable {
-					if t.currentNode != node && t.changed != nil {
+					previousNode := t.currentNode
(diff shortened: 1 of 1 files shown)


millerlogic:
My use case is:
I want to be able to use treeview nodes as "links" to other primitives, but the treeview sets focus after calling the selected handler, so I [can't focus](https://github.com/rivo/tview/pull/431/commits/089ac467f881a0f21b0161b5aba54f5d1ee148a0) my "linked" primitive because it'll go right back to the treeview.
Also in some cases I want to clear or change the currently selected node immediately when selected, but it can result in the wrong handler being called or even a panic (consider if ```t.currentNode``` is changed in the first selected handler [here](https://github.com/rivo/tview/pull/431/commits/77ccc1ff41099c94ee1fbf6615bb1bf7e1524750))
Thanks

rivo:
Looks ok to me, thanks.

PR #434, opened 2020-04-19
Add ContextMenu

tslocum:
Initially supported by List


ka1r03:
@rivo, any update on this? Also, can we reuse dropdown widget here?

ka1r03:
Does this work with mouse (right-click to show the context menu)?

PR #437, opened 2020-04-24
Add List.Transform and TreeView.Transform

tslocum:


tslocum:
This change is available now in [cview](https://gitlab.com/tslocum/cview), a fork of tview.

rivo:
Hmm, this is not really my preferred style. I don't think `Transform` is the right name for something like this and I'm not sure an extra type should be added when there's only little benefit. If needed at all, I will probably take a different route here.

Issue #64, opened 2018-03-05
Feature request: textarea


uthng:
Hi,

I have just found your package. It is very interessting and I am trying to write an application with it. 
I would like to know if there is any plan to create a **texarea** in the near future ?

Thanks.
NT

rivo:
Hi, glad you like it. I don't have concrete plans for a text area yet. I have been thinking about extending the `TextView` class to make it editable. However, adding that would get me very close to implementing a full text editor:

- Adding/removing characters
- Navigating the text (not trivial with automatic line breaks etc.)
- Selecting areas, cutting/pasting (optional at first maybe)
- Deal with `TextView`'s color/region tags somehow
- Rewriting the `Form` class so editable `TextView` objects can also be included (e.g. this could break the horizontal layout)

So it's not a minor thing to add. I could add an "on roadmap" label to this issue if this is something you definitely need. This would mean I will eventually add it. But it could then still be some time before that happens and I don't know how urgent your need is here. (See #23 for an example of an "on roadmap" feature.)

Do you have a specific use case where an `InputField` is not enough?

uthng:
I have intention to use tview to write several console applications. And in some of them, I need to modify a value under JSON format for example. I understand that it is not easy to but I think it is necessary when we edit some complexe data such as json, html etc. I have been thinking about making TextView to be editable. But I have not really entered in the code yet.

rivo:
For your own project, you could probably come up with something simple by using `TextView`, intercepting key events with [`SetInputCapture()`](https://godoc.org/github.com/rivo/tview#Box.SetInputCapture), and using [regions and highlights](https://godoc.org/github.com/rivo/tview#TextView) to simulate a cursor.

`TextView` is already very complex. You can try to extend it but with so many things to consider -- and given that I have to maintain it in the long term -- I'm not sure I will merge any pull requests in that direction.

So I'm adding the "on roadmap" label. This means I will implement this but I don't know yet when. I will have to dedicate quite some time to it.

uthng:
Yes, I think that I can manage in my projects so dont worry about when you
will implement it. Anyway, thanks to add "on roadmap" for this request
because it will be really nice to have it as a base element.

-------------
Thanh NGUYEN

glycerine:
[...]

PR #447, opened 2020-05-14
TextArea

Konstantin8105:
Create for #64
* Adding/removing characters
* Navigating the text
* Graphemes support
* Test for finding corner cases
* Add `demos` source
* Minimal change into `TextView`
* No color
* No regions
* No `Form` interface


textarea.go
```diff
@@ -0,0 +1,465 @@
...
+				f.cursorByBuffer(line, pos)
+			}()
+		case tcell.KeyBackspace, tcell.KeyBackspace2:
+			line, pos = f.deleteRune()
+		default:
+			r := event.Rune()
+			if unicode.IsLetter(r) ||
+				unicode.IsNumber(r) ||
+				unicode.IsMark(r) ||
+				unicode.IsSpace(r) {
```
berrange, line 421: This filtering is too strict - it blocks user from entering any special characters, eg !"£$%^&*(){}~@:<>?  and more. I think it could be reduced to just  "unicode.IsGraphic", so that it only blocks control sequences.

jpeletier:
@Konstantin8105 Given the pace at which reviews go in this repo, I'd suggest you move this widget to your own staond-alone repo under your control so we can begin using this and you can get PRs for improvements. I did this with my widget, [winman](https://github.com/epiclabs-io/winman) so it would be available immediately to everybody.

textarea.go
Konstantin8105: I agree with you.

Konstantin8105:
@jpeletier  In according to LICENSE, you can merge that widget in your project without any link on myself. Feel free to coping, using, and ...

jpeletier:
@Konstantin8105 Yes, thank you for a permisive license, but I would like to avoid copying your code if I can better link to it and everyone shares the same to improve it together.

This repo is quite "slow". If you keep this in your repo, we can probably evolve your TextArea widget faster together.

xxwassyxx:
This is actually a very very useful addition. Is there any reason it cannot be merged into master. Just trying to get a picture on timeline as this is also what I need.

buzzdan:
watching this one as well...

skanehira:
This is very useful!
I waiting and I hope TextArea added

l13t:
Also waiting for merge of this PR.

brlbil:
I have been waiting for this future to start a personal project. So I appreciate it if it can be merged asap.

rivo:
I know this will be disappointing to some of you but it's unlikely I will merge this PR. There are a whole bunch of reasons but the two major reasons are: I don't think `TextArea` should be based on `TextView` because it would inherit all the shortcomings of `TextView`, mainly its inability to handle large texts as well as its non-optimal performance. The second major reason is that this PR has a bug at its core that will not be trivial to fix: Runes are not Unicode graphemes. A grapheme can consist of multiple runes. (A lot of emojis, e.g. flags, consist of multiple runes.) This will lead to bugs; I've already seen and fixed many related bugs.

Please fill out #578, `TextArea` is one of the features you can "vote" for. (So far, no one has voted for the text editing component. But there's still time to fill out the questionnaire.) If it comes out on top, it will be one of the next big things I will tackle.

textarea.go
```diff
@@ -0,0 +1,465 @@
...
+	f.cursor.y = indexLine + y - f.view.lineOffset
+	if isOutsideBuffer {
+		f.cursor.x++
+	}
+	if y+height <= f.cursor.y {
+		diff := (indexLine - lastIndexLine) - (y + height - lastCy) + 1
+		f.view.lineOffset += diff
+	}
+	if f.cursor.y < y {
+		diff := -((lastIndexLine - indexLine) - lastCy) - 1
```
berrange, line 352: As best I can tell 'lastCy' shouldn't be involved in this calculation at all.  When the textarea fills the screen lastCy will be zero when we hit the top of the screen and need to scroll earlier lines into view.  When the textarea is offset from the top of the screen though, lastCy will be non-zero and this will cause `diff` to get ridiculous values. As a result lineOffset gets set to incorrect values and all your text disappears off the textarea display. Removing 'lastCy' from this calc made scrolling work as expected.

rivo:
`TextArea` exists now. (Since last year already.)

Issue #192, opened 2018-11-23
Disabling and hiding form fields


dimonomid:
First of all, thanks a lot for your work on tview, I'm using it in a small project and so far it works quite well. There are just a few missing pieces, one of which is: disabling and hiding form fields. E.g. depending on the other values of the form, it often makes sense to disable/enable some input fields, or even hide/show them.

I tried to mimic it by setting a different background color for the field I want to disable, but that didn't work either because the `Form` component overrides it by calling `SetFormAttributes` for all components. So far I just set text to `n/a` and use acceptance function which stops accepting any values when the field should be disabled, but that's far from perfect anyway: the field should change its appearance to make it clear that it's disabled, and should't receive focus.

Do you think you'd be open to implementing that?

rivo:
There is [`RemoveFormItem()`](https://godoc.org/github.com/rivo/tview#Form.RemoveFormItem) which allows you to remove a form item. But, granted, it's not the same as "hiding" it as you will have to reinsert it somehow.

And you are correct, there is no way to disable a form item. I'm not even sure what's a good way to make a field look like it's disabled - and it will have to work for all field types as well, i.e. input fields, checkboxes, drop-downs, and buttons. Currently, `tview`'s default colour scheme only uses primary colours as more may not be available on some systems. So it's a bit more difficult to make it work well using colours.

But I understand your use case. I'm certainly open to adding it. (As my time is limited, please don't expect this to happen immediately.)

If you have any ideas on how to display disabled form items so it's clear to the user that they are disabled, let me know.

rivo:
Any thoughts?

dimonomid:
Hi @rivo , yeah sorry for not replying earlier, I was hoping to actually implement something myself and then show to you.

> If you have any ideas on how to display disabled form items so it's clear to the user that they are disabled, let me know.

I think just using the gray font on a black background would make it clear that the field is disabled. But I'd have to try it out first, and play with the colors.

rivo:
Keep in mind that some terminals have very few colours like white, red, yellow, blue, green, cyan, magenta. My own terminal does not diplay "gray".

dimonomid:
Good point, I guess we'd have to stick to just white on black, and with dashes displayed instead of the value. So e.g. while the field is enabled, it displays white `0.0` on blue background (as it does now), and when it's disabled, it just displays white `---` on black background, and isn't focusable.

rivo:
[...]

PR #449, opened 2020-05-27
Add support for read only form items

berrange:
When a form item is readonly, any input events or mouse events that
would cause its value to change are ignored.

Read only items are rendered when text color that matches the label
color to distinguish them from items that accept input.

This implements half of the feature request in #192


berrange:
An alternative impl was merged long ago in [47b3275](https://github.com/rivo/tview/commit/47b3275db48de684be18375672ea5c5671cad6e2)

PR #452, opened 2020-05-31
TextView: Add SetMaxLines()

### LINKED ISSUE #451: TextView: limiting the maximum number of lines
Hi, 

I am using `TextView` to display scrolling log output, but as soon as the text in the buffer becomes very big, CPU usage shoots up.

I propose solving it by adding a new `SetMaxLines()` method that makes sure the buffer slice never contains more lines than the set maximum. By default, it is set to infinite, therefore this would be a backward-compatible change.

Proposed PR: #452

jpeletier:
Fixes #451 

This PR implements a new method `TextView.SetMaxLines()` that limits the amount of lines that can be added to a buffer. This is useful to use TextView to show scrolling content, such as log output.

If this method is not used, the default is "no limit", thus this method is backward-compatible and does not affect existing behavior of the TextView

### DIFF
--- a/textview_test.go
+++ b/textview_test.go
@@ -1,0 +1,57 @@
+package tview_test
+
+import (
+	"fmt"
+	"strings"
+	"testing"
+
+	"github.com/rivo/tview"
+)
+
+func TestTextViewMaxLines(t *testing.T) {
+	tv := tview.NewTextView()
+
+	// append 100 lines with no limit set:
+	for i := 0; i < 100; i++ {
+		_, err := tv.Write([]byte(fmt.Sprintf("L%d\n", i)))
+		if err != nil {
+			t.Errorf("failed to write to TextView: %s", err)
+		}
+	}
+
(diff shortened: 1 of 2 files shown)


textview_test.go
```diff
@@ -0,0 +1,57 @@
...
+
+	// append 100 more lines:
+	for i := 100; i < 200; i++ {
+		_, err := tv.Write([]byte(fmt.Sprintf("L%d\n", i)))
+		if err != nil {
+			t.Errorf("failed to write to TextView: %s", err)
+		}
+	}
+
+	// Sice max lines is set to 20, we should still get 20 lines:
```
ghostsquad, line 44: type in `Since`

jpeletier: Fixed. Thanks

rivo:
Please see #451 for the reasons for closing this PR.

rivo:
Refer to https://github.com/rivo/tview/issues/451#issuecomment-780116886 for a more detailed explanation.

PR #480, opened 2020-07-26
Fix bug in list RemoveItem

zmnpl:
When a changed function is set, deleting the first element (0-index) lead to a index out of range error because the change event was fired with index = -1.
Now the index will only be decreased if the last item was removed. As a side effect that changes the behavior of the selection slightly:
It will stay at the same place and only move up when the last item was removed.

### DIFF
--- a/list.go
+++ b/list.go
@@ -148,9 +148,9 @@ func (l *List) RemoveItem(index int) *List {
 		return l
 	}
 
-	// Shift current item.
+	// Shift current item when it was the last.
 	previousCurrentItem := l.currentItem
-	if l.currentItem >= index {
+	if l.currentItem >= len(l.items) {
 		l.currentItem--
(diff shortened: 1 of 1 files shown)


zmnpl:
Hi, any optinions on this? It is a two lines fix after all :)

Regards

nyanco01:
Hello.
I am currently suffering from the same bug. Please fix it as I believe it can be fixed with a few changes.

Best regards.

rivo:
The latest commit should fix this.

If you want issues like this to be resolved sooner (or at all), please read the [Contributing Guidelines](https://github.com/rivo/tview/blob/master/CONTRIBUTING.md), specifically the section about opening an issue before submitting a pull request. (Especially since the proposed solution of this PR would have introduced more bugs.)

PR #497, opened 2020-09-07
Added new project to README

ezeoleaf:
Hi, I just added a project I recently released that uses tview. Thanks!

Feel free to take a look :)


modal.go
```diff
@@ -95,6 +95,17 @@ func (m *Modal) SetText(text string) *Modal {
 	return m
 }
 
+/// AddInputText adds buttons to the window. There must be at least one button and
```
mushkevych, line 98: Following docklet might be better suited:

```suggestion
// AddInputText adds Input Fields to the window, which are later on identified by their labels.
```

ezeoleaf: Looks good, thanks. Updating it.

PR #396, opened 2020-02-01
WIP Add scroll bar to List

tslocum:
Support is implemented for DropDown and InputField.  Vertical scroll support is implemented for Table, I would appreciate assistance with adding horizontal scroll support.  I am also seeking feedback generally on this PR.

PR #513, opened 2020-10-13
Add support for horizontal scrolling to List

### LINKED ISSUE #512: Horizonal scroll for List
I want to add horizonal scrolling to the list element, because for wide tables not everything could fit the screen.

image

There is no setting for this on the List element https://pkg.go.dev/github.com/rivo/tview#List

What is the proper way to do this? An example in wiki would be nice to have.

tslocum:
Resolves #512.

This change is available in [cview](https://gitlab.com/tslocum/cview), a fork of tview.

### DIFF
--- a/list.go
+++ b/list.go
@@ -54,8 +54,9 @@ type List struct {
 	// Whether or not navigating the list will wrap around.
 	wrapAround bool
 
-	// The number of list items skipped at the top before the first item is drawn.
-	offset int
+	// The number of list items and columns by which the list is scrolled
+	// down/to the right.
+	itemOffset, columnOffset int
 
 	// An optional function which is called when the user has navigated to a list
 	// item.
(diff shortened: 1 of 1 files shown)


abitrolly:
Is there an indicator that a column is scrolled?

tslocum:
No horizontal indicator is drawn.  I [suggested changes](https://github.com/rivo/tview/pull/396) adding vertical scroll indicators in January, which could be adapted for this.

abitrolly:
@tslocum maybe it is possible to modify List to be easier extendable to add these features in custom UI rather than patching and maintaining custom version of `tview`?

tslocum:
Please see [this document](https://gitlab.com/tslocum/cview/blob/master/FORK.md) for an explanation of why I forked tview and the differences between the two.

rivo:
@tslocum Just curious, have you tried this with multi-rune Unicode characters or even with characters that occupy more than one cell?

rivo:
Well, the proper solution took me a few hours to figure out. It's not as easy as it may have seen because, as in many other places,

```
byte ≠ rune ≠ Unicode character
```

So I can't merge this PR. Sooner or later someone gets problems with emojis or foreign characters.

Issue #521, opened 2020-11-05
Form border lines are rendered white on white since tcell update


berrange:
If I build the 'demos/form/main.go' demo program and run, the form used to have a double line border around it with background in black and lines in white, along with the title text in white.

Since the update to tcell in 

```
commit e4d167311d7708471e449a05b3d1c9dfb061abfc (HEAD)
Author: Oliver <oliver@stockperformer.com>
Date:   Sun Oct 18 14:15:57 2020 +0200

    Upgraded to latest tcell version. Results in a minor incompatibility in the Table class.```

The form now does not render the border correctly, both the background and lines and the text are all in white.

**Comment by rivo** [comment]
I cannot reproduce this with the latest version:

![image](https://user-images.githubusercontent.com/480930/99435686-c36f5100-2910-11eb-8018-a05d2cf87b33.png)

Which terminal / OS are you using?

**Comment by berrange** [comment]
This is with latest tview source, cloned right now, and then building from the git root

```
$ go build demos/form/main.go 
go: downloading github.com/rivo/uniseg v0.2.0
go: extracting github.com/rivo/uniseg v0.2.0
go: downloading github.com/gdamore/tcell/v2 v2.0.1-0.20201017141208-acf90d56d591
go: extracting github.com/gdamore/tcell/v2 v2.0.1-0.20201017141208-acf90d56d591
go: downloading golang.org/x/sys v0.0.0-20201017003518-b09fb700fbb7
go: downloading golang.org/x/text v0.3.3
go: extracting golang.org/x/sys v0.0.0-20201017003518-b09fb700fbb7
go: extracting golang.org/x/text v0.3.3
go: finding github.com/gdamore/tcell/v2 v2.0.1-0.20201017141208-acf90d56d591
go: finding golang.org/x/text v0.3.3
go: finding golang.org/x/sys v0.0.0-20201017003518-b09fb700fbb7
go: finding github.com/rivo/uniseg v0.2.0
$ ./main
```

I'm using Fedora 33, x86_64, with GNOME Terminal  (gnome-terminal-3.38.1-1.fc33.x86_64).

Normally I'm using a white-ish background, and black text, but for this screenshot I've set my terminal background to pink, to make it obvious that the borders are inheriting the terminal's default background colour instead of any tview colour theme:

![tview-borders](https://user-images.githubusercontent.com/1517315/99510960-93ab6200-297f-11eb-99f2-6a0903767180.png)

joegrasse:
[...]

Issue #519, opened 2020-11-02
Modal border background color no longer matches modal's background color


polothy:
The modal border background color used to match the background color of the modal itself.  Now it appears that the border background color is the default background color.

I think this change introduced the new behavior: https://github.com/rivo/tview/blob/d551c850a7435cbb7829ce6703fd470c664801f3/box.go#L73

On the modal, the border is set on the `frame` which is private, so cannot workaround it by just updating the border style:

https://github.com/rivo/tview/blob/d551c850a7435cbb7829ce6703fd470c664801f3/modal.go#L49-L51

This PR could help to provide a way to implement a workaround: https://github.com/rivo/tview/pull/458 - but that's just an FYI. I of course do not mind how it is fixed or if it is actually implemented as intended and this is not a bug.

Since I'm making an issue, I'll take the opportunity to thank you for the amazing project! Been having a lot of fun using it 😄

mevdschee:
I have also seen this happen and I would appreciate a fix.

polothy:
![presentation_demo](https://user-images.githubusercontent.com/634657/98141788-e3683480-1e7b-11eb-94c7-29bcdb7ad026.png)

You can see an example of the coloring in the presentation demo on the Flex and Grid demo pages.

mevdschee:
@polothy Thank you, I see exactly the same. And @rivo thank you for creating this wonderful library. I really like building stuff with it!

rivo:
The latest commit should fix this. Could you please give me feedback whether it works for you now?

(Sorry for the delay, btw.)

mevdschee:
![image](https://user-images.githubusercontent.com/1288217/110970899-7f3e0b80-835a-11eb-9c16-f4a59fe3fc8a.png)

Works on my machine! Thank you.

rivo:
More fixes. (Please refer to https://github.com/rivo/tview/issues/520#issuecomment-797659408.) Does it work for you with the latest update?

mevdschee:
>Does it work for you with the latest update?

I tried the "Hello world again" code and it failed previously, but works fine for me when upgraded from:

    github.com/rivo/tview v0.0.0-20201018122409-d551c850a743 => v0.0.0-20210312174852-ae9464cc3598

I hope this of use to you. Kind regards, Maurits

rivo:
Great! Thanks!

PR #523, opened 2020-11-05
Apply background colour when drawing Box border

berrange:
The borderStyle field only encodes the foreground colour of the border,
leaving the background to inherit from the terminal default theme. This
often means the border is draw white-on-white, which is a regression
introduced by the tcell v2 update.

It needs to honour the background colour set for the Box instance
when drawing the border.

Fixes: https://github.com/rivo/tview/issues/521

### DIFF
--- a/box.go
+++ b/box.go
@@ -337,6 +337,8 @@ func (b *Box) Draw(screen tcell.Screen) {
 		}
 	}
 
+	borderStyle := b.borderStyle.Background(b.backgroundColor)
+
 	// Draw border.
 	if b.border && b.width >= 2 && b.height >= 2 {
 		var vertical, horizontal, topLeft, topRight, bottomLeft, bottomRight rune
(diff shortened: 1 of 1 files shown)


polothy:
Closed https://github.com/rivo/tview/pull/524 in favor of this PR.  This one looks better and it fixes the same problem.  You can see before/after screenshots of the fix in https://github.com/rivo/tview/pull/524.

polothy:
This PR would also fix https://github.com/rivo/tview/issues/519

splicemaahs:
I experienced the same issue, and applied this patch locally to test, it did fix the odd display problem.

problame:
Can confirm that this would help.

Before

image


After

image

problame:
(Notably, we don't set background colors anywhere. I just noticed this issue when updating from a ~1y old state of tview to latest).

1 year old state:

image

mevdschee:
@problame Can you confirm that the latest version has this fix and close this PR?

problame:
We switched to `cview` in the meantime :/

berrange:
Closing as it was fix in commit ae9464cc3598d81fff6813af3e28f292a9de6c15

PR #560, opened 2021-02-05
Add configurable color for Input autocomplete

issadarkthing:
It would be convenient if we can configure the color of the autocomplete for `InputField` instead of using global `Styles`.

### DIFF
--- a/inputfield.go
+++ b/inputfield.go
@@ -213,6 +213,42 @@ func (i *InputField) SetAutocompleteFunc(callback func(currentText string) (entr
 	return i
 }
 
+// SetAutocompleteMainTextColor sets the autocomplete main text color.
+func (i *InputField) SetAutocompleteMainTextColor(color tcell.Color) *InputField {
+	if i.autocompleteList != nil {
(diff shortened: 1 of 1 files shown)


gpanders:
I'd like to add my support for this feature. The inability to style the autocomplete list separately from the rest of the primitives is fairly limiting, particularly when trying to support both light and dark terminal backgrounds.

gpanders:
@issadarkthing I actually don't think this will work. In the `Autocomplete` function, the list is destroyed and re-created repeatedly:

https://github.com/rivo/tview/blob/d4fb0348227b252e5ec51b4ad8426c9e63646822/inputfield.go#L232-L248

Instead, you either need to 1) add a new hook that can be called for the autocomplete list to be configured after it's created or 2) add new properties to InputField (`autocompleteTextColor`, `autocompleteBackgroundColor`, etc.) and use those to set the autocomplete list styles in lines 243-247.

I have a PR ready to go that does the latter: https://github.com/gpanders/tview/commit/c5d98ee108aaf795b639029c77fffa1d0f2dcdaa

PR #566, opened 2021-02-10
Add some keys to move cursor in InputField

skanehira:
I added some keys  likes Emacs style to move cursor in InputField


skanehira:
This key was already added

PR #598, opened 2021-05-11
Add flag selectedReverseColor to reverse style of selected item

mdom:
I'm working on xterm with disabled color support. This patch allows me to see the currently selected item in a list. This is probably not a good solution, there are surley be other widgets with similar problems. But i just wanted to see if the reverse code worked.

What do you think about adding a api, where i can just set styles directly for elements? Like _SetSelectedStyle_, _SetBorderStyle_, etc?


rivo (changes requested):
Thanks for this PR. This is generally a good idea and one step closer to resolving #611. I have a few review comments and requests for changes. Please have a look.

PR #610, opened 2021-05-28
Add support for styling the borders used with modals.

gdamore:
It turns out that I need to have more control over the styling of these borders.  This small addition fixes that.  (Note that due to inheritance from Box we had this method already, but it basically didn't do what you expected -- we need to control the frame's borders, not the box's.

### DIFF
--- a/modal.go
+++ b/modal.go
@@ -77,6 +77,18 @@ func (m *Modal) SetButtonTextColor(color tcell.Color) *Modal {
 	return m
 }
 
+// SetBorderColor sets the color of the border.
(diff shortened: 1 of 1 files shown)


gdamore:
This is now so old that it is probably obsolete.  I'm closing it to remove it from my radar -- I'm no longer using the tview package for my daily work anyway.

rivo:
On a side note, styling the borders (colours etc.) is something that I'm working on at the moment. Will look quite a bit different from this proposal, though.

PR #622, opened 2021-06-17
add emacs-like key

skanehira:
Add emacs-like key for selecting autocomplete list

### DIFF
--- a/inputfield.go
+++ b/inputfield.go
@@ -598,13 +598,13 @@ func (i *InputField) InputHandler() func(event *tcell.EventKey, setFocus func(p 
 			} else {
 				finish(key)
 			}
-		case tcell.KeyDown:
+		case tcell.KeyDown, tcell.KeyCtrlN:
 			if i.autocompleteList != nil {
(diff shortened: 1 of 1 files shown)


aditya-K2:
@rivo I think we can expose a function for user like SelectNext or SelectPrevious so that the user can map the keys on his own end.

aditya-K2:
any updates on this?

rivo:
You can use [SetInputCapture](https://pkg.go.dev/github.com/rivo/tview#Box.SetInputCapture) to define your own key binds. I don't see why this PR is needed, especially because `tview` does not implement emacs-like key binds anywhere else.

aditya-K2:
> You can use [SetInputCapture](https://pkg.go.dev/github.com/rivo/tview#Box.SetInputCapture) to define your own key binds. I don't see why this PR is needed, especially because `tview` does not implement emacs-like key binds anywhere else.

`SetInputCapture` captures the input for the `input field` and not the autocomplete menu. How can one access Autocomplete's Input?

Issue #636, opened 2021-08-19
panic when scrolling textview


SamWhited:
When calling `ScrollTo` on a textview I'm frequently getting the following (trimmed to relevant parts) panic starting in the textview's draw function at line 1017:

```
$ go run .
panic: runtime error: index out of range [100] with length 0 [recovered]
        panic: runtime error: index out of range [100] with length 0

goroutine 1 [running]:
github.com/rivo/tview.(*Application).Run.func1()
        /home/sam/go/pkg/mod/github.com/rivo/tview@v0.0.0-20210624165335-29d673af0ce2/application.go:243 +0x4d
panic({0x946f80, 0xc00059b200})
        /home/sam/gotip/src/runtime/panic.go:1038 +0x215
github.com/rivo/tview.(*TextView).Draw(0xc0000bc100, {0xa43700, 0xc000568000})
        /home/sam/go/pkg/mod/github.com/rivo/tview@v0.0.0-20210624165335-29d673af0ce2/textview.go:1017 +0xdf4
```

rivo:
Please see https://github.com/rivo/tview/wiki/Concurrency. Most of the functions are not thread-safe. You need to take care of that yourself. (The `Write()` function is an exception and I'm not sure if, for consistency reasons, it should've been like the other functions.)

SamWhited:
This is all happening synchronously and there should be no race possible here

rivo:
I don't see your code so it's hard to say. Do you write to the `TextView` using the `io.Writer` interface, calling `Clear()` periodically? And other parts of the program could cause a non-main-thread redraw. Is this a correct assumption?

SamWhited:
I think so, I haven't managed to get this into an easily reproducible form, and it's been long enough that I don't really remember what was happening now. It is possible (though unlikely, I think?) that other areas of the code could be calling `app.Redraw()` which could be trickling down to this.

SamWhited:
I have tested this, I still get the panic even with locking around the scrolling.

rivo:
Responding to https://github.com/rivo/tview/pull/641#issuecomment-925438772 here: This is actually what I would like to suggest. In fact, if it weren't for backwards compatibility, I'd probably even remove the lock from the `Write()` function and leave everything up to the user.

When I implemented this, I was under the impression that implementing the `io.Writer` interface required thread safety so I introduced a mutex. Now it's a bit inconsistent, leading to the confusion outlined in this issue. That's my bad.

[...]

PR #641, opened 2021-09-09
Make TextView.Clear safe

sachaos:
Fix https://github.com/rivo/tview/issues/636

I think the problem is TextView.Clear is not locking.
When we clear the TextView, the buffer will be nil.
So call Clear while drawing buffer, it will panic with that stack trace.

```
github.com/rivo/tview.(*TextView).Draw(0xc0000bc100, {0xa43700, 0xc000568000})
        /home/sam/go/pkg/mod/github.com/rivo/tview@v0.0.0-20210624165335-29d673af0ce2/textview.go:1017 +0xdf4
```

### DIFF
--- a/textview.go
+++ b/textview.go
@@ -419,6 +419,8 @@ func (t *TextView) GetScrollOffset() (row, column int) {
 
 // Clear removes all text from the buffer.
 func (t *TextView) Clear() *TextView {
+	t.Lock()
+	defer t.Unlock()
 	t.buffer = nil
 	t.recentBytes = nil
 	t.index = nil


rgrannell1:
+1, I've encountered similar issues that were resolved by locking the `TextView`

```go
ctx.tgt.Lock()
ctx.tgt.Clear()
ctx.tgt.Unlock()
```

SamWhited:
ooooh, the TextView itself is lockable, I didn't realze that's something I could do. That will almost certainly fix it, thanks.

SamWhited:
I wanted to quickly follow up and request a different strategy: instead of merging this, let's remove the locking from around writes and leave it to the user. Otherwise if I want to clear a buffer and then do a bunch of writes I incur lock overhead each time. Instead I would prefer to lock at the beginning of my function, clear, do writes in a tight loop, etc. then unlock.

rgrannell1:
if you take that approach, please document it clearly in the concurrency section of the wiki!

PR #652, opened 2021-09-23
Add batch writer for TextView's

SamWhited:
I'd like to propose this as an alternative superset to #641 (See also discussion in #636).

This allows for backwards compatibility by not modifying the locking characteristics of `Write`, but also gives us a way to perform writes in a tight loop without incurring the overhead of locking.

Usage would look like this:

```go
tv := tview.NewTextView()

w := tv.BatchWriter() // Lock aquired
defer w.Close() // Lock released
w.Clear()
for {
  w.Write(/* some long write, copy, etc. in a loop */)
}
```


textview.go
```diff
@@ -419,10 +419,17 @@ func (t *TextView) GetScrollOffset() (row, column int) {
 
 // Clear removes all text from the buffer.
 func (t *TextView) Clear() *TextView {
+	t.Lock()
```
SamWhited, line 422: This part is the same as  #641 so that PR can be merged first if desired.

textview.go
```diff
@@ -634,9 +646,6 @@ func (t *TextView) Write(p []byte) (n int, err error) {
 		}()
 	}
 
-	t.Lock()
```
SamWhited, line 637: Part of the slowness I had been experiencing was that there are actually two locks here, but they don't appear to be necessary (it's just skipping over a single branch, might as well keep it locked unless we expect / want things to change between checking if changed should be called and doing the actual write, but I don't see why we'd want that)

PR #666, opened 2021-10-30
Add "alternate bold" to list widget

Pauloo27:
Hello.

I want to suggest a feature to the list widget: "alternating bold" in the list items to improve readability. 

Without it:
image

With it:
image


rivo:
I think people can use colour tags if they need this. I don't feel that this needs to be part of the base functionality. I'm not even sure if it renders correctly on all the terminals. It also doesn't play well with secondary text, in my opinion.

Issue #671, opened 2021-11-14
Please Add Functionality to set BackgroundColor For autocomplete. 


aditya-K2:
The [Default](https://github.com/rivo/tview/blob/badfa0f0b301e121ab4c072ffb4bdc81451541f0/inputfield.go#L280) Background color doesn't look good. Having a method to change the Autocomplete menus background would be useful.

rivo:
There is now a function `InputField.SetAutocompleteStyles()` that allows you to change the styles of the autocomplete drop-down.

Issue #611, opened 2021-05-28
Would be nice to expose tcell.Style directly


gdamore:
Many of the widgets here use tcell styles, but the tview API only exposes the Color explicitly.  This leads to inconsistencies in some places.  For example, the border can usually have attributes set on it, but other things can't.  Some things can have their background color changed, and others can't.

I have a use case where I'd like to be able to provide underlining via attributes to things (basically I want to use List with colors and an underbar to make it look like sort of like hyperlinks).

It  turns out that making Attributes available everywhere is also possibly very important, because in some settings we are not permitted to use color exclusively to convey meaning.  (Due to color blindness).  Using attributes would let us change the way the text is rendered to add underbars, reverse text (which *can* be simulated with colors), blink, italics, etc.

Essentially, I'd like *every* widget to have methods:

SetStyle(), SetBorderStyle(), SetTitleStyle(), etc.

rivo:
It's just a historical artifact that `tview` uses `Color` more than `Style`. I guess when I started the project, I thought colours would be all people needed. But I was obviously wrong.

I can look into adding additional functions to set styles. But it has to be backwards compatible.

golangdojo99:
+1

rivo:
I'm closing this as there's no need to track this issue anymore. I have added a lot of methods to set/get the `tcell.Style` object directly already and continue to add them where they are still missing.

PR #672, opened 2021-11-15
Customize Colors of Autocomplete List 

aditya-K2:
As I mentioned here #671 

Added the Functionality to Customize the following colors of the
Autocomplete list :

	1. Main Text
	2. Selected Background Color
	3. Selected Text Color
	4. Background Color

The Colors can be set through the Following Functions

	1. SetAutocompleteMainTextColor
	2. SetAutocompleteSelectBackgroundColor
	3. SetAutocompleteSelectedTextColor
	4. SetAutocompleteBackgroundColor

### DIFF
--- a/inputfield.go
+++ b/inputfield.go
@@ -76,6 +76,18 @@ type InputField struct {
 	autocompleteList      *List
 	autocompleteListMutex sync.Mutex
 
+	// Main Text Color of the Autocomplete List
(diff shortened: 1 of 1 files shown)


aditya-K2:
Should I also add get methods too?

aditya-K2:
I don't think this implementation will work

rivo:
See also #611 as to why I've opted to implement this differently.

PR #680, opened 2021-12-19
Adding podman-tui and tvxwidgets to the list of projects that use tview

navidys:


navidys:
- Adding podman-tui to the list of projects that use tview
- Adding tvxwidgets to the list of projects that use tview

Issue #692, opened 2022-01-25
Feature request: focus border colors


kylehereagain:
Border color for focused widgets/views.

rivo:
The double lines are not enough?

syncjuncture:
> The double lines are not enough?

This would be an awesome addition. This is just an opinion, but double lines look rather ugly.

PR #712, opened 2022-03-15
Add borderFocusStyle to box

icechen128:
The simple addition of a `borderFocusStyle` addresses the issues mentioned in #692. If there is a better solution or other considerations, please let me know.


NikSays:
Would also be nice to add `BorderFocusColor` default value to styles.go, and use it in `NewBox` definition, so that it can be set for every primitive. if @icechen128 is no longer active, I can do it

PR #719, opened 2022-04-03
Fix AddDropDown

### LINKED ISSUE #718: Form: AddDropDown immediately calls SelectedFunc
Hi,

When you add a new DropDown item to a form, the "selected" func is immediately executed. (because the handler is registered prior to the SetCurrentOption call)

This does not seem to happen for any of the other form items.

regards,
MO

---
Very nice lib btw. I love it :wink:

moson-mo:
SelectedFunc should be assigned after setting the CurrentOption.
Otherwise the function is already called when the DropDown is being added.

Fixes #718


moson-mo:
Might cause problems with applications that rely on the current behaviour.

PR #739, opened 2022-06-29
Add logic to flush channels, and to safeguard adding of events, after…

bhbosman:
Add logic to flush channels, and to safeguard adding of events, aftermain loop in Application.Run exited

### DIFF
--- a/application.go
+++ b/application.go
@@ -688,18 +688,19 @@ func (a *Application) draw() *Application {
 // the screen.
 func (a *Application) Sync() *Application {
 	// check to see if the Application.Run is still valid
+	msg := queuedUpdate{
+		f: func() {
+			a.RLock()
+			screen := a.screen
+			a.RUnlock()
+			if screen == nil {
+				return
+			}
+			screen.Sync()
+		},
+	}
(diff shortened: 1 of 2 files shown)


digitallyserviced:
Thanks for this!!! 

I just started trying out this library and after seeing the done guard application.go:380 I honestly have no idea what the point of QueueUpdate{,Draw} was because the there was never anything setup to be on the other side of the channel to pump something so it can drain... and because it relies on a message it received rather than the status of the channel, closing it is useless too... That change was made 13 months ago... has it been broken that long?

bhbosman:
Hi, 

From the examples, and the general use case, the `tview/Application` will always be in the main method and will cease to exist, when the application terminate. 

In my use case, I am using  the `tview/Application` to manage a reader/writer stream from a `SSH` channel. This is a bit different from the general use case, and my application does not terminate when the `SSH` /  `tview/Application` terminate. I need to close and drain the channels, to clear resources. 

I removed some of the done guards, as it was excessive. Can not answer you on `QueueUpdate{,Draw}`

Issue #14, opened 2018-01-11
How do I navigate the presentation demo program


pjebs:
I can get to section 7 but can't proceed

triddell:
Issue #10 has a hint. The author "actually built in a short cut: Hit Ctrl-N to advance to the next slide, Ctrl-P for the previous slide."

rivo:
I've added that hint to the presentation itself. Hopefully, that will clear things up.

PR #748, opened 2022-07-26
Add another word wrap function allowing own patterns

moson-mo:
There might be cases where you want to define your own pattern for splitting lines.

This commit adds "WordWrapWithPattern" which accepts a regexp pattern.

Signed-off-by: moson-mo <mo-son@mailbox.org>

### DIFF
--- a/util.go
+++ b/util.go
@@ -460,17 +460,16 @@ func stringWidth(text string) (width int) {
 }
 
 // WordWrap splits a text such that each resulting line does not exceed the
-// given screen width. Possible split points are after any punctuation or
-// whitespace. Whitespace after split points will be dropped.
+// given screen width. Split points are defined by a regular expression
 //
 // This function considers color tags to have no width.
 //
 // Text is always split at newline characters ('\n').
-func WordWrap(text string, width int) (lines []string) {
+func WordWrapWithPattern(text string, width int, pattern *regexp.Regexp) (lines []string) {
 	colorTagIndices, _, _, _, escapeIndices, strippedText, _ := decomposeString(text, true, false)
 
 	// Find candidate breakpoints.
-	breakpoints := boundaryPattern.FindAllStringSubmatchIndex(strippedText, -1)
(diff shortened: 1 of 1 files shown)


rivo:
Package [`rivo/uniseg`](https://github.com/rivo/uniseg) has been updated to include line breaking (word wrapping) according to [Unicode Standard Annex #14](https://unicode.org/reports/tr14/) (UAX 14). I haven't updated all classes / functions in `tview` to use this new code yet but I will, including this `WordWrap()` function. After that, we can reevaluate if this change is needed. (Using regexes will go away so this PR will have to be modified.)

It would also be good for you to let me know more about your use cases, i.e. in which context do you need a different word wrapping function and how is it different from the default algorithm or UAX 14?

Additionally, since `WordWrap()` is a simple helper function with no effects on the `tview` package, what keeps you from simply implementing your own word wrapping function if you need it? Why does that have to go into `tview` if it's not being used by the package?

moson-mo:
> It would also be good for you to let me know more about your use cases, i.e. in which context do you need a different word wrapping function and how is it different from the default algorithm or UAX 14?

I don't want to split on "-" since it is a part of the "keyword".

> Additionally, since WordWrap() is a simple helper function with no effects on the tview package, what keeps you from simply implementing your own word wrapping function if you need it? Why does that have to go into tview if it's not being used by the package?

Nothing.

moson-mo:
whoooops. accidentally closed this,

moson-mo:
> Using regexes will go away

Oh, ok. Closing this then.

rivo:
> I don't want to split on "-" since it is a part of the "keyword".

I'm not sure what "keyword" you're referring to but it might interest you that UAX#14 has the following rules:

> Do not break before hyphen-minus, other hyphens [...]. 
> Do not break within ‘——’, even with intervening spaces.

It might break _after_ a hyphen but not for numbers (e.g. "$-9.23").

Issue #751, opened 2022-08-03
Textview can't get GetRegionText with colored content and panics


sruehl:
There is a guard to only check if the current tag `currentTag` has a index in `colorTagIndices`: 
https://github.com/rivo/tview/blob/711ef394f9b31ad219f5f7bbc85bddfb766af6ab/textview.go#L623
when it hits the end tag it increments the `currentTag`
https://github.com/rivo/tview/blob/711ef394f9b31ad219f5f7bbc85bddfb766af6ab/textview.go#L625
and then produces a panic in the next line because the new tag is not bound checked against `colorTagIndices`
https://github.com/rivo/tview/blob/711ef394f9b31ad219f5f7bbc85bddfb766af6ab/textview.go#L627

sruehl:
so from my understanding that would be solved by changing 627 to `if currentTag < len(colorTagIndices) && colorTagIndices[currentTag][1]-colorTagIndices[currentTag][0] > 2 { `

PR #753, opened 2022-08-05
fix: fixed GetRegionText when colors are active

sruehl:
This PR fixes https://github.com/rivo/tview/issues/751

When skipping color tags the `currentTag` is getting incremented and directly after that a index access with that happens. This always crashes on the last tag.

Without this fix `GetRegionText` is unusable when color is active


sruehl:
@rivo this is a pretty simple fix. Would be great if that could be merged as without colors can't be used in conjunction of `GetRegionText`

rivo:
I agree this is a bug so thank you for spotting it.

But your commit doesn't solve this issue or, said differently, it introduces another bug. Line 627 should never use the incremented `currentTag` value. If you simply add a bounds check, `continue` will not execute when it should, namely for the last colour tag which is not empty (`[]`).

It should look like this:

```go
tag := currentTag
if pos == colorTagIndices[tag][1]-1 {
	currentTag++
}
if colorTagIndices[tag][1]-colorTagIndices[tag][0] > 2 {
	continue
}
```

If you modify your PR accordingly, I'll merge it.

sruehl:
Oh sorry, somehow one of my github orgs spams so much notifications that your message didn't reach me. I will fix that ASAP.

I force-pushed your suggestions.

Issue #734, opened 2022-06-20
New tview based app


aurc:
Hello,

Just released an app based on tview/tcell and I was wondering it could be listed with the apps on your page. 
https://github.com/aurc/loggo

Thanks
Aurelio

rivo:
Please leave a pull request with your change for the README file and I'll merge it.

aurc:
Thanks! Done: A terminal app for structured log streaming (GCP stack driver, k8s, local streaming)

PR #755, opened 2022-08-11
Update README.md - added loggo app on projects using tview

aurc:
Listed loggo in the projects that are using tview in the README.md: A terminal app for structured log streaming (GCP stack driver, k8s, local streaming)


aurc:
Ref: https://github.com/rivo/tview/issues/734

rivo:
Look nice, btw!

PR #758, opened 2022-08-20
vimMove Variable in TextView

Awesome-Sauces:
This is to block the actions preformed when these keys are pressed in a textview
- h, left arrow: Move left.
- l, right arrow: Move right.
- j, down arrow: Move down.
- k, up arrow: Move up.
- g, home: Move to the top.
- G, end: Move to the bottom.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -1,4 +1,4 @@
-module github.com/rivo/tview
+module github.com/Awesome-Sauces/tview
 
 go 1.18
 
--- a/go.mod
+++ b/go.mod
@@ -1,4 +1,4 @@
-module github.com/Awesome-Sauces/tview
+module github.com/rivo/tview
 
 go 1.18
 
--- a/textview.go
+++ b/textview.go
@@ -193,6 +193,10 @@ type TextView struct {
 	// navigated when the text is longer than what fits into the box.
 	scrollable bool
 
+	// If set to true, the text view will navigate similarly to vim.
+	// If set to false, vim movement will be ignored.
+	vimMove bool
+	
 	// If set to true, lines that are longer than the available width are wrapped
 	// onto the next line. If set to false, any characters beyond the available
 	// width are discarded.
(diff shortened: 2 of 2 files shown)


rivo:
Why?

If you want different key binds (or none), you can use [`SetInputCapture()`](https://pkg.go.dev/github.com/rivo/tview#Box.SetInputCapture).

Awesome-Sauces:
Well in my mind I find it easier to just call a function to stop those key binds from moving you around in the textview but still allow scrolling at the same time.

rivo:
I'm sorry but you haven't provided a reason yet why offering these key binds is a problem.

Next someone comes and says they only want Vim bindings but no arrow movement.
Then someone comes along and says they need the same flag for all the other primitives, too.

It's precisely the reason why I'm offering `SetInputCapture()` so everyone can customize it to their own preferences.

So unless there's a good reason to offer this specific option to all users of `tview`, it's going to be a "no" from my side.

Awesome-Sauces:
Oh ok would you mind explaining shortly how SetInputCapture would work for me being able to ignore the Vim bindings? I can't really find a way to ignore the Vim bindings with SetInputCapture.

rivo:
Sure, no problem. Here's an example:

```go
func main() {
	app := tview.NewApplication()
	textArea := tview.NewTextView().SetText(text)
	textArea.SetInputCapture(func(event *tcell.EventKey) *tcell.EventKey {
		switch event.Key() {
		case tcell.KeyLeft, tcell.KeyRight, tcell.KeyUp, tcell.KeyDown:
			return event // Allow arrow keys.
		}
		return nil // Ignore all others.
	})
	if err := app.SetRoot(textArea, true).EnableMouse(true).Run(); err != nil {
		panic(err)
	}
}
```

Awesome-Sauces:
Oh thanks! I can't believe it was as simple as just returning nil :(

rivo:
No problem. I'm glad I was able to help!

Issue #797, opened 2023-01-22
Window freeze if selecting textview


dingxiong:
Hey guys, I encounter a problem with textview in a form. The window freezes after I try to select a text view.

The tview version I am using is `v0.0.0-20230104153304-892d1a2eb0da`. The code is 
```
package main

import (
        "github.com/rivo/tview"
)

func main() {
        app := tview.NewApplication()
        form := tview.NewForm().
                AddDropDown("Env", []string{"production"}, 0, func(option string, index int){

                }).
                AddInputField("env", "", 20, nil, nil).
                AddInputField("Last name", "", 20, nil, nil).
                AddTextArea("Address", "", 40, 0, 0, nil).
                AddTextView("Notes", "This is just a demo.\nYou can enter whatever you wish.", 40, 2, true, false).
                AddCheckbox("Age 18+", false, nil).
                AddPasswordField("Password", "", 10, '*', nil).
                AddButton("Save", nil).
                AddButton("Quit", func() {
                        app.Stop()
                })
        form.SetBorder(true).SetTitle("Enter some data").SetTitleAlign(tview.AlignLeft)
        if err := app.SetRoot(form, true).EnableMouse(true).Run(); err != nil {
                panic(err)
        }
}
```
<img width="407" alt="Screenshot 2023-01-21 at 7 46 41 PM" src="https://user-images.githubusercontent.com/5496402/213897039-a8ccc07d-32ff-41db-bed2-353c6a1241ae.png">

I tried to select the the notes fields `This is just a demo.` multiple times, then the UI freezes. By selection, I mean I put the mouse cursor on the message, and press the muse left button, and scroll to the right. 

Below is the debug info.
```
$ dlv attach 44630
Type 'help' for list of commands.
(dlv) grs
  Goroutine 1 - User: /opt/homebrew/Cellar/go/1.19.2/libexec/src/runtime/sema.go:77 sync.runtime_SemacquireMutex (0x100d31984) [semacquire]
  Goroutine 2 - User: /opt/homebrew/Cellar/go/1.19.2/libexec/src/runtime/proc.go:364 runtime.gopark (0x100d08f74) [force gc (idle)]
  Goroutine 3 - User: /opt/homebrew/Cellar/go/1.19.2/libexec/src/runtime/proc.go:364 runtime.gopark (0x100d08f74) [GC sweep wait]
[...]

PR #798, opened 2023-01-27
fix deadlocking nonscrollable textview #797

digitallyserviced:
Resolves the issue with the textviews when set non-scrollable in a form causing a deadlock with the focus handlers set by the form.

Since the `t.scrollable=false` seems to mean the textview is read-only plus no scrolling the mouse handler should return earlier so no setFocus handlers run.

#797 should be resolved by this.


rivo:
`scrollable` does not mean the text view is read-only. This PR has unwanted side-effects, e.g. highlighting regions in non-scrollable text views would not work anymore. #797 has been solved in a different way.

PR #799, opened 2023-01-28
Add vertical/horizontal radio button group as a form item

vendelin8:
Also add a radio group to `form` demo


vendelin8:
I forgot to replace the form demo screenshot, but my colors are somewhat different from the screenshots. So I think the best would be if you recreate it. Thanks.

PR #816, opened 2023-02-23
Introduce stopOnCtrlC in application.

gizahNL:
This allows consuming applications signal that they intend to keep running when control+c is pressed, and instead want the event to be handled by their primitives.

My use case is for a window in which I run shell commands; And that I want to cancel via pressing control+c. The current default of stopping on control+c is kept so there should not be any surprises for anyone updating.

### DIFF
--- a/application.go
+++ b/application.go
@@ -83,6 +83,9 @@ type Application struct {
 	// Set to true if mouse events are enabled.
 	enableMouse bool
 
+	// Set to true if application should stop when control + C is pressed
+	stopOnCtrlC bool
+
 	// An optional capture function which receives a key event and returns the
 	// event to be forwarded to the default input handler (nil if nothing should
 	// be forwarded).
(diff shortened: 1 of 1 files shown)


rivo:
This is what [`Application.SetInputCapture()`](https://pkg.go.dev/github.com/rivo/tview#Application.SetInputCapture) is for. This specific use case is actually documented there. You can also use it to bind "Close Application" to a different key.

gizahNL:
> This is what [`Application.SetInputCapture()`](https://pkg.go.dev/github.com/rivo/tview#Application.SetInputCapture) is for. This specific use case is actually documented there. You can also use it to bind "Close Application" to a different key.

Can you perchance explain how to do it that way? I am able to block handling of control+c that way completely, but I don't see a way to let control+c be passed down chain without the application getting stopped.

rivo:
Actually, your question made me realize that there was indeed no way to forward Ctrl-C to primites. You were able to block it or translate it but not forward.

I fixed this now. There is now a specific way to achieve this. Check out the link above for details.

gizahNL:
> Actually, your question made me realize that there was indeed no way to forward Ctrl-C to primites. You were able to block it or translate it but not forward.
> 
> I fixed this now. There is now a specific way to achieve this. Check out the link above for details.

Many thanks! That will work for me !

PR #836, opened 2023-04-07
adding an option to disable dropdown primitive automatic selection of an item

navidys:
Adding an option to disable dropdown primitive automatice selection of an item based on prefix.
Some applications using vim like key binding such as:
- j: Move down
- k: Move up
- l: Move right
- h: Move left

Dropdown primitive by default will select an input based on input key rune (prefix) that will cause issue with such a key bindings. Its possible to change to parse and change the key even in the application side, however I thought it will be more cleaner to be have an option.

### DIFF
--- a/dropdown.go
+++ b/dropdown.go
@@ -71,6 +71,10 @@ type DropDown struct {
 	// possible.
 	fieldWidth int
 
+	// Set true when automatic seletion of an item in the drop-down list
+	// based on the prefix is enabled (default is true).
+	autoSelection bool
+
 	// An optional function which is called when the user indicated that they
 	// are done selecting options. The key which was pressed is provided (tab,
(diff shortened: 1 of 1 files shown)


navidys:
Hi @rivo 
Any comment regarding this PR ?

Regards

rivo:
> Its possible to change to parse and change the key even in the application side, however I thought it will be more cleaner to be have an option.

Cleaner maybe for your application but more clutter for this package. I'm tending towards not merging this and suggesting that you use `SetInputCapture()` to change the behaviour if you prefer vim-like navigation here.

I might change my mind if multiple people ask about this or if someone can demonstrate that the current implementation is a real problem. But for now, I'd say I'll keep it the way it is.

PR #848, opened 2023-05-10
Fix TreeView CollapseAll

ViRb3:
Fixes a typo that causes CollapseAll to collapse the root node multiple times instead of its descendants.


rivo:
Thanks!

PR #874, opened 2023-08-28
added Page.GetPageNames(visible_only bool) API

carpii:


carpii:
Addresses the limitation that there's no way to iterate over pages currently, which is useful for highly dynamic apps (and providing Alt-Tab functionality)

GiGurra:
This would be great to have. Is there any way to upvote/get this some attention?

Or is the project semi dead? :(

rivo:
See https://github.com/rivo/tview/blob/master/CONTRIBUTING.md#pull-requests

But looking at this now, this PR has multiple issues:

- Using snake case for variable names
- Using the `slices` package which requires Go 1.21. We're at Go 1.18 in `tview` and last time I upgraded the minimum version (and I did so long after Go 1.18 was released), I was showered with tons of complaints because it broke a lot of builds. The `slices` package has been introduced officially about half a year ago so I would like to wait a bit before bumping the Go version again.

carpii:
Pull request replaced with https://github.com/rivo/tview/pull/970

rivo:
#970 looks good but you closed it...

carpii:
Sorry, I made a real meal of it 

Replaced with #971

PR #889, opened 2023-09-26
Fix grammar in description of SetColumns method

paololazzari:


grid.go
```diff
@@ -77,8 +77,8 @@ func NewGrid() *Grid {
 
 // SetColumns defines how the columns of the grid are distributed. Each value
 // defines the size of one column, starting with the leftmost column. Values
-// greater 0 represent absolute column widths (gaps not included). Values less
-// or equal 0 represent proportional column widths or fractions of the remaining
+// greater than 0 represent absolute column widths (gaps not included). Values less
+// than or equal to 0 represent proportional column widths or fractions of the remaining
```
rivo, line 81: Can you keep lines to a maximum of 80 characters, please?

rivo (changes requested):

grid.go
paololazzari: Fixed!

Issue #896, opened 2023-10-04
Add func (*TreeNode) GetParent 


paololazzari:
Being able to get the parent of a given node would make traversing the tree much easier

paololazzari:
Then `func (*TreeNode) AddChild` could automatically set the parent of the child, and it would then be possible to implement a `func (*TreeNode) GetChildren` function

paololazzari:
@rivo I have something similar to the demo here https://github.com/rivo/tview/tree/master/demos/treeview . When I press enter and a node which corresponds to a valid file is selected, I want to open that file in a textview. Opening the file for files within folders requires having the name of the parent nodes too

rivo:
Isn't your actual requirement then to get the entire path to the current node, not just to the parent? Like the breadcrumbs you see on many websites?

If so, a better solution would then be to provide a `GetPath(node *TreeNode) []*TreeNode` method which gives you all nodes leading to your current node, starting at the root.

What do you think?

paololazzari:
Yes, absolutely!

paololazzari:
Actually, this seems to be simple enough already. As long as the reference is set to the full path (as shown [here](https://github.com/rivo/tview/blob/master/demos/treeview/main.go#L30) then its trivial to get the full path of the selected now with `GetCurrentNode().GetReference().(string)`

rivo:
Well, I added the function (check out the latest commit) while I was travelling and offline so I only saw your message afterwards. But you are right, in the demo, the full path is stored as a reference object so it's available anyway.

digitallyserviced:
@paololazzari 

Just my $1.39...

As far as being able to get the parent of a tree's node, I agree with @rivo in that this is not something you should be reliant upon due to the transient nature the parent of a given leaf node is.

While it may seem odd that parent is temporary, or should not be relied upon for what you are attempting is the fact that the parent/root node of a tree, can change all the time. 

Also if you try and rely on what should only be GUI/view objects, for handling the representation of an underlying tree/graph such as a filesystem or hierarchy, you're gonna have a bad time. If you are going to actually try using it to determine the state/position of the actual data structure represented, you will run into problems. 

The fact is that the tree's ***View/GUI/Nodes/Leafs/Left Leg/Religion*** can be refreshed/re-rendered accurately at any time from the tree/graph it represents.

The same really is not true for the opposite. 

[...]

PR #897, opened 2023-10-04
Add func (n *TreeNode) GetParent()

paololazzari:
Fixes https://github.com/rivo/tview/issues/896

### DIFF
--- a/treeview.go
+++ b/treeview.go
@@ -143,6 +143,11 @@ func (n *TreeNode) RemoveChild(node *TreeNode) *TreeNode {
 	return n
 }
 
+// GetParent returns this node's parent.
+func (n *TreeNode) GetParent() *TreeNode {
+	return n.parent
+}
+
 // SetSelectable sets a flag indicating whether this node can be selected by
 // the user.
 func (n *TreeNode) SetSelectable(selectable bool) *TreeNode {


rivo:
The problem with this is that `parent` is a temporary variable (as you can see in its declaration comment) which is set during traversal of the tree. That means, it is not guaranteed to be set. For example, it will not be set for nodes that have not been traversed. It will also not be set if the tree has not been drawn yet. And it might even be wrong if you make changes to the tree structure before drawing it.

The correct implementation for a `GetParent` function is to walk the tree until you've found the node. Then you'll know the parent and can return it. This would work but it would be expensive, especially for large trees, so this fact must be described in the function comments.

If you want to provide such an implementation, feel free to submit one. As for the current PR, I don't think it's a good idea to return a temporary variable with almost no guarantees.

paololazzari:
Thank you for the quick response @rivo 

Why is `parent` a temporary variable though? why not have it be a normal variable? this could easily be set for the root node and for the other nodes it could be set when `AddChild` is called on them

rivo:
I guess I should have anticipated this question. This is an architectural choice, to keep the number of variables that need to by synchronized as low as possible. It's the same with `tview` in general: Widgets like `Flex` or `Grid` will know their contained widgets but they will not know who contains them.

It's too easy to make mistakes and end up with an inconsistent data structure if references go both ways.

You might say "but it's quite easy to do it here", to which I'd say "it might be now but given that I have maintained this code for years and will maintain it for many years to come, I cannot be sure that a future change will lead to issues".

But as I said above, `GetParent` is possible and I would prefer that solution.

PR #927, opened 2023-12-07
fix remove dropdown option index out of range

7yyo:
@rivo 

Hi, when setSelectFunc of dropdown to remove current option, if the last option is selected, index out of range will panic. I tried to fix this bug.

The code below will reproduce this bug if select the last option `Fifth`.

```go
package main

import "github.com/rivo/tview"

func main() {
	app := tview.NewApplication()
	dropdown := tview.NewDropDown()
	dropdown.SetLabel("Select an option (hit Enter): ")
	selectFunc := func(text string, index int) {
		dropdown.RemoveOption(index)
	}
	dropdown.SetOptions([]string{"First", "Second", "Third", "Fourth", "Fifth"}, selectFunc)
	if err := app.SetRoot(dropdown, true).EnableMouse(true).Run(); err != nil {
		panic(err)
	}
}

```

### DIFF
--- a/dropdown.go
+++ b/dropdown.go
@@ -518,13 +518,10 @@ func (d *DropDown) openList(setFocus func(Primitive)) {
 		}
 		if len(d.options) != 0 {
 			if d.currentOption >= len(d.options) {
-				if d.options[d.currentOption-1].Selected != nil {
-					d.options[d.currentOption-1].Selected()
-				}
-			} else {
-				if d.options[d.currentOption].Selected != nil {
-					d.options[d.currentOption].Selected()
-				}
+				d.currentOption--
+			}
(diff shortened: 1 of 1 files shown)


rivo:
Thanks for catching this. Two things needed to be done here:

1. After line 516, store `d.options[d.currentOption]` in a local variable, e.g. `currentOption`. Use it in the lines that follow (3 times). This block is just to call the "selected" callbacks, the one for all options and the one defined by one option. Your solution could lead to the wrong callback to be called because you're adjusting `d.currentOption` before calling the option's "selected" callback.
2. Set `d.currentOption` to -1 in `RemoveOption` if the removed option is the current one. Document this in the function comments.

I ended up making these changes myself. I hope you don't mind.

PR #960, opened 2024-03-22
Try to fix #954

### LINKED ISSUE #954: Undesirable List Behavior
For the following code:
```go
	list := tview.NewList().
		AddItem("hello", "", 0, nil ).
		AddItem("World","" , 0, nil ).
		AddItem("!","" , 0, nil ).
		AddItem("qwertyiopasdfghjkl;","" , 0, nil ).
		ShowSecondaryText(false).
		SetHighlightFullLine(true).
		SetSelectedStyle(tcell.StyleDefault.Reverse(true).Foreground(tcell.ColorBlue))
```

The rendered list looks like this:

image

It seems that the selected style with the reversed attribute set to true is only being applied to the `mainText` portion of the list, I would like the given style to be applied to the full line instead of just the `mainText`

Souheab:
This PR attempts to fix #954

### DIFF
--- a/list.go
+++ b/list.go
@@ -525,9 +525,10 @@ func (l *List) Draw(screen tcell.Screen) {
 		// Background color of selected text.
 		if index == l.currentItem && (!l.selectedFocusOnly || l.HasFocus()) {
 			textWidth := width
+			taggedStringWidth := TaggedStringWidth(item.MainText)
 			if !l.highlightFullLine {
-				if w := TaggedStringWidth(item.MainText); w < textWidth {
-					textWidth = w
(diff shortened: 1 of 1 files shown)


rivo:
The "print text first, then reverse colours on selections" method was still a leftover from the first version of this package. As you correctly noted, this failed in some cases. But the fix should not have been to keep this (IMHO bad) method around and patch it somehow but to use styles correctly by printing the selected text with the "selected style", then (optionally) extending the line with that same style.

By working on this, I also came across a few other bugs in the package which I fixed here as well.

PR #985, opened 2024-05-08
Add deadlock guard for Queue{Update,Event}

diamondburned:
This commit adds an atomic bool into Application that ensures that the
user can never call QueueUpdate or QueueEvent in the main loop.

Previously, when the user does this, the application would deadlock and
freeze forever. The user would often have to kill the process manually
to get out of this state. Now, the application will panic, indicating
the bug immediately.

### DIFF
--- a/application.go
+++ b/application.go
@@ -3,6 +3,7 @@ package tview
 import (
 	"strings"
 	"sync"
+	"sync/atomic"
 	"time"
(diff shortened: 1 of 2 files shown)


diamondburned:
This method is buggy and won't work as expected. Closing for now.

PR #1002, opened 2024-06-25
Add "oddshub" to "Projects using `tview` in README.md

dos-2:


rivo:
image

Btw, `tview` requires at least Go 1.18.

PR #1007, opened 2024-07-15
Feat: add GetSelectedFunc method for TreeView component

### LINKED ISSUE #1004: Feature request: add `GetSelectedFunc` method for `TreeView` component
I've created wrappers for some library components those change wrapped component's default behavior on input. This follows a decorator pattern. My logic handles some inputs while relying on the component's default behavior for others. I understand I could create my own components to achieve this but I'd like to keep doing it that way.

Most library components have a comprehensive interface. All setters have corresponding getters, for example, `SetDrawFunc`-`GetDrawFunc`, `SetInputCapture`-`GetInputCapture`, etc. However, some components do **not** follow this pattern. For instance, `TreeView` has `SetSelectedFunc` but lacks `GetSelectedFunc`.  

Here's a minimal example of the wrapper I've created
```go
type SubtreeDeleteWrapper struct {
	deletedNodes map[*tview.TreeNode]struct{}
}

func (w *SubtreeDeleteWrapper) Wrap(t *tview.TreeView) *tview.TreeView {
	// get a wrapped tree view's default input capture handler
	defaultCapture := t.GetInputCapture()

	// add some custom input capture for wrapped tree view
	t.SetInputCapture(func(event *tcell.EventKey) *tcell.EventKey {
		switch event.Rune() {
		case 'D': // delete a selected node and its subtree
			selectedNode := t.GetCurrentNode()
			if _, exists := w.deletedNodes[selectedNode]; exists {
				break
			}
			// some business logic goes here e.g. delete an entry in a database
 			// or only register that deletion so it could be canceled later
			
			// just collapse the selected node's subtree and mark it as deleted
			// but it could be something different, maybe more complicated
			selectedNode.SetExpanded(false)
			selectedNode.SetColor(tcell.ColorRed)
			w.deletedNodes[selectedNode] = struct{}{}
		case 'Z': // undo deletion
			selectedNode := t.GetCurrentNode()
			if _, exists := w.deletedNodes[selectedNode]; !exists {
				break
			}
			// ...
			selectedNode.SetColor(tcell.ColorGreen)
			delete(w.deletedNodes, selectedNode)
		default:
			// otherwise, call the default input capture handler
			if defaultCapture != nil {
				return defaultCapture(event)
			}
			return event
		}
		return nil
	})

	// here I want to get a tree view's default `SelectedFunc` but I can't
	var defaultSelectedFunc func(*tview.TreeNode) = nil // ???

	t.SetSelectedFunc(func(node *tview.TreeNode) {
		// if the selected node is deleted, do nothing on that selection
		if _, exists := w.deletedNodes[node]; exists {
			return
		}
		// otherwise call the default selection handler
		if defaultSelectedFunc != nil {
			defaultSelectedFunc(node)
		}
	})

	return t
}
```

and I use it that way
```go
func main() {
	tree := tview.NewTreeView()
	// add tree nodes, etc
	// ...

	tree = util.NewSubtreeDeleteWrapper().Wrap(tree)
	// initialize and run an app
}
```

alexey-savchuk:
Close #1004

### DIFF
--- a/treeview.go
+++ b/treeview.go
@@ -450,6 +450,12 @@ func (t *TreeView) SetSelectedFunc(handler func(node *TreeNode)) *TreeView {
 	return t
 }
 
+// GetSelectedFunc returns the function set with SetSelectedFunc() or nil
+// if no such function has been set.
+func (t *TreeView) GetSelectedFunc() func(node *TreeNode) {
+	return t.selected
+}
+
 // SetDoneFunc sets a handler which is called whenever the user presses the
 // Escape, Tab, or Backtab key.
 func (t *TreeView) SetDoneFunc(handler func(key tcell.Key)) *TreeView {
--- a/treeview.go
+++ b/treeview.go
@@ -450,8 +450,8 @@ func (t *TreeView) SetSelectedFunc(handler func(node *TreeNode)) *TreeView {
 	return t
 }
 
-// GetSelectedFunc returns the function set with SetSelectedFunc() or nil
-// if no such function has been set.
+// GetSelectedFunc returns the function set with [TreeView.SetSelectedFunc]
+// or nil if no such function has been set.
 func (t *TreeView) GetSelectedFunc() func(node *TreeNode) {
 	return t.selected
 }


treeview.go
```diff
@@ -450,6 +450,12 @@ func (t *TreeView) SetSelectedFunc(handler func(node *TreeNode)) *TreeView {
 	return t
 }
 
+// GetSelectedFunc returns the function set with SetSelectedFunc() or nil
```
rivo, line 453: Please use the Go link syntax here, i.e.

> ... set with [TreeView.SetSelectedFunc] or ...

You might have to change the line wrapping to adhere to the 80 character limit.

rivo (changes requested):

treeview.go
alexey-savchuk: Done.

PR #1024, opened 2024-09-13
Add SelectItem to list view

### LINKED ISSUE #1023: Add ability to select list item programatically
Right now in an application I have a list of items. Some of the items can be selected with a keyboard shortcut other than `KeyCR` (let's say '1' selects the first item, for the sake of this example). Right now to select the item I have to do something like this (in an `InputHandler()` implementation, for the sake of this example, but if you were using `SetInputHandler` you'd have to return the fake event and it's broadly similar):

```
    if event.Rune() == '1' {
      myList.SetCurrentItem(0)
      myList.InputHandler()(tcell.NewEventKey(tcell.KeyCR, 0, tcell.ModNone), nil)
    }
```

It would be nice to have a way to either get the "selected" function from the current list item, or a function to select an item just as if enter were pressed without having to manually trigger a keyboard event. This would make the example much easier to read:

```
    if event.Rune() == '1' {
      myList.SelectItem(0)
    }
```

I'm happy to implement this if you like the API and need a hand.

*EDIT:* realized this is a bad example because if it were just '1' you could use the shortcut rune, but in my actual case it's a more complicated key combo and this won't work. Anyways, keep in mind that it's just an example.

Also, went ahead and made a PR just to make it easy to see what the change would be like. Happy to change the API or close it if it doesn't get accepted, of course.

SamWhited:
This allows the user to programatically call an items select function as if the Enter key had been pressed without requiring the use of fake key signals that act as a sort of "shadow" API on top of the actual Go API.

Fixes #1023

### DIFF
--- a/list.go
+++ b/list.go
@@ -101,6 +101,19 @@ func NewList() *List {
 	}
 }
 
+// SelectItem calls the given item's select function (if defined).
+func (l *List) SelectItem(index int) {
+	if index >= 0 && index < len(l.items) {
+		item := l.items[index]
+		if item.Selected != nil {
+			item.Selected()
+		}
+		if l.selected != nil {
+			l.selected(index, item.MainText, item.SecondaryText, item.Shortcut)
+		}
+	}
+}
+
 // SetCurrentItem sets the currently selected item by its index, starting at 0
 // for the first item. If a negative index is provided, items are referred to
 // from the back (-1 = last item, -2 = second-to-last item, and so on). Out of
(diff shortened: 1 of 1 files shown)


list.go
```diff
@@ -101,6 +101,19 @@ func NewList() *List {
...
+		item := l.items[index]
+		if item.Selected != nil {
+			item.Selected()
+		}
+		if l.selected != nil {
+			l.selected(index, item.MainText, item.SecondaryText, item.Shortcut)
+		}
+	}
+}
+
```
ccoVeille, line 116:
```suggestion
func (l *List) SelectItem(index int) {
	if index < 0 || index >= len(l.items) {
		return
	}
	
	item := l.items[index]
	if item.Selected != nil {
		item.Selected()
	}
	if l.selected != nil {
		l.selected(index, item.MainText, item.SecondaryText, item.Shortcut)
	}
}

```


But maybe this could simply work

```suggestion
func (l *List) SelectItem(index int) {
	item, found := l.items[index]
	if !found {
		return
	}
	if item.Selected != nil {
		item.Selected()
	}
	if l.selected != nil {
		l.selected(index, item.MainText, item.SecondaryText, item.Shortcut)
	}
}

```

SamWhited: I generally agree with your stylistic choices on the code, but I was trying to keep the diff as minimal as possible (ie. just moving what the original author already wrote and not re-write their code and introduce the possibility of mistakes introducing new and unexpected behaviors). Happy to change it if the maintainer wants though, I do prefer yours and this is simple enough.

SamWhited:
Gentle ping; never heard from a maintainer on this.

rivo:
Closing this, see [#1023](https://github.com/rivo/tview/issues/1023#issuecomment-2453398295). I think `SelectItem` would not have been an optimal name because it wouldn't have selected an item. Anyway, I think the current solution should solve your problem.

Issue #1031, opened 2024-10-08
Colors don't reset when using Bolt or Italic


sruehl:
Colors don't reset when using Bolt or Italic from https://github.com/fatih/color for example.
Here is a testcase demonstration the current issue:
```golang
package tview

import (
	"github.com/fatih/color"
	"testing"
)

func TestTranslateANSIBroken(t *testing.T) {
	info := color.New(color.FgGreen, color.Bold).Sprint
	info2 := color.New(color.FgGreen, color.Italic).Sprint
	color.NoColor = false
	greenText := info("this should be colored green and bolt")
	t.Log(greenText)
	greenText2 := info2("this should be colored green and italic")
	t.Log(greenText2)
	t.Log(TranslateANSI(greenText + " uncolored " + greenText2))
}

```

![image](https://github.com/user-attachments/assets/16b6ce42-d464-430a-907b-de60c14e975b)

the offending line is this one `if len(params) == 0 || len(fields) == 0 && fields[0] == "0" {`

With the above code corrected the output is like that:
![image](https://github.com/user-attachments/assets/d1df01a0-9d3b-49b7-85aa-67763be8cfa6)

rivo:
Thanks. The latest commit fixes this and some other issues in the ANSI translation code.

PR #1032, opened 2024-10-08
fix: color should now properly reset when using bolt or italic

sruehl:
fixes https://github.com/rivo/tview/issues/1031


sruehl:
@rivo quick little fix for broken colors when using bolt or italic. For more info see #1031

PR #1036, opened 2024-10-14
Remove abandoned project from readme

SamWhited:
I got a bit excited to see this while looking through the projects in the readme, but it looks like all it ever did was import tview, then no one ever wrote any code and it was abandoned 5 years ago.


rivo:
Thanks!

PR #1055, opened 2024-11-23
Fix docs about multi-line text in forms

SamWhited:
Previously TextAreas could not be added to forms. This functionality has been added, but the documentation still said that it was not possible.


SamWhited:
Gentle ping

Issue #330, opened 2019-07-31
Feature request: dropdown multi-select


qjcg:
With forms, it would be great to be able to allow users to choose multiple selections from a DropDown.

Examples of this feature:

- <https://semantic-ui.com/modules/dropdown.html#multiple-selection>
- <https://ant.design/components/select/#components-select-demo-multiple>
- <https://blueprintjs.com/docs/#select/multi-select>

In tview, perhaps when a multi-select dropdown is opened, elements could each have a CheckBox area next to them (ex: `[ ]`) and hitting space on a dropdown entry might toggle its selection (`[X]`).

rivo:
Thanks for this suggestion. I don't see this on my immediate todo list but I'll keep this here for future development.

owen-carter:
i want need this

7yyo:
+1

manuelacgu:
+1

PR #1058, opened 2024-11-23
Add multi-select mode for lists

SamWhited:
This adds a multi-select mode for lists in which the current selection changes to be a cursor and pressing enter marks the item under the cursor as actually selected.

DropDown is based on list so this should unblock adding a multi-select mode to DropDown's as well.

Updates #330


SamWhited:
Something else I thought of after preparing this that I'd like to get your opinion on: for multi-select in drop down's we'll need a callback for when an option is unselected as well as selected. We could re-use `SetSelectedFunc` and pass in a negative index to indicate that it was unselected. Eg. an index of 1 means the second item is selected, an index of -1 would mean that the second item was unselected. Because this would only be applicable when multi-select is turned on, current programs would continue to work as expected.

SamWhited:
Gentle ping; no rush of course, I know you're busy.

JensvandeWiel:
Would love this to be added

Gorynychdo:
Is it possible to add an exportable method to get the list of selected items?

SamWhited:
Definitely seems like a good idea, though probably a separate feature / follow up would be a better place to do that? It doesn't seem like there's interest in merging this anyways, so I'm not sure that I'd hold out hope.

PR #1105, opened 2025-07-26
Fix race conditions in TreeView

roelvandergoot:
When you:
1.Represent a directory tree as a `TreeView`,
2. Use a state machine go-routine per leaf `TreeNode`, and
3. Append the current state name of the state machine to the `TreeNode`'s using `TreeNode.SetText()`,
a race condition exists between `TreeNode.SetText()` and `TreeView.Draw()`

You can't solve the race condition by embedding `rivo.TreeView` and `rivo.TreeNode`s in your own `TreeView` and `TreeNode`s. Hence this pull request.


roelvandergoot:
Thank you for pointing me to that wiki page. I was not aware of the `app.QueueUpdate` and `app.QueueUpdateDraw` functions. With them I should be able to prevent the race conditions.

Oooh! `rivo/tview` is getting semantic versions! Thank you again.

